PR #17, opened 2015-10-07
Extract dropdown into a standalone component and compose over it

cibernox:
Extracted dropdown functionality into https://github.com/cibernox/ember-basic-dropdown

This new component is in charge of:

* Inject the wormhole destination in the body.
* Render in the wormhole and position absolutely calculating the coordinates (or render in place)
* Add/remove scroll/resize/orientation events and position the dropdown above or below the trigger.
* Also uses MutationObservers (with fallback in older browsers) to detect when the content of the dropdown changes and reposition again.
* Basic a non select-specific behavior like toggle when the trigger is clicked/pressEbter, close when clicked outside it, close on escape and disable itself (and remove tabindex)

That leaves the select component with less responsibility and ~100 LOC less :)

There is however a small hack in this approach. The ideal API would have been

```hbs
{{#ember-basic-dropdown ...options as |toggleDropdown|}}
  {{!-- Here you can pass the toggleDropdown action as a regular argument to some actions 
  so the select component can open and close it --}}
{{else}}
  {{!-- Sadly the yielded argument is not available in the inverse block --}}
{{/ember-basic-dropdown}}
```

Therefore the small hack on `didInsertElement` I inject the dropdown component in the select component (https://github.com/cibernox/ember-power-select/blob/extract_dropdown_component/addon/components/ember-power-select.js#L147-L149) so I can call open/close directly on it.

Since nothing prevents in ember-land to yield arguments to the inverse block i'll try to push the htmlbars story on this.

/cc @miguelcobain


cibernox:
Finally I managed to yield the action and use that as public API. Injecting the component was too hacky.

PR #40, opened 2015-10-14
Create bootstrap theme

cibernox:
Created bootstrap theme.

For now only Sass users can use it, by doing:

```scss
// app.scss
@import 'ember-power-select/themes/bootstrap';
@import 'ember-power-select';
```

Solution for css users soon :)


miguelcobain:
I think we're not quite there yet:

- [ ] The inner input doesn't have border-radius
- [ ] Multiple selection doesn't show the blue focus box-shadow
- [ ] Normal bootstrap inputs have an inset box shadow (`box-shadow: inset 0 1px 1px rgba(0,0,0,.075);`)
- [ ] Input and buttons have a fixed default height of `34px`. I imagine having inputs and power-selects side-by-side, it would be strange if they don't have the same height.
- [ ] Dropdown should have a padding of `5px` besides the `2px` margin that separates it from the trigger.
- [ ] Each dropdown item is also different. They have a height of 26px in bootstrap, it seems. Also paddings are different.
- [ ] Dropdown item hover color is brighter in bootstrap (`#f5f5f5`)
- [ ] Option group headers should use bootstrap's dropdown headers: http://getbootstrap.com/components/#dropdowns-headers
- [ ] Caret down is different
- [ ] In multiple selection, I feel the label color for each selected option is too dark, but I didn't find a bootstrap color for it yet.

PR #41, opened 2015-10-14
add blockless form. closes #28

### LINKED ISSUE #28: Add blockeless form
Not completely sold because it doesn't add any value other tab being more concise for very simple use cases, but still worth considering

miguelcobain:


addon/templates/components/ember-power-select/multiple.hbs
```diff
@@ -8,7 +8,11 @@
     {{#if resultsLength}}
       {{#ember-power-select/options options=results highlighted=_highlighted selection=selection
         highlight=(action "highlight") select=(action "select") dropdown=dropdown as |option|}}
-        {{yield option}}
+        {{#if baseHasBlock}}
+          {{yield option}}
+        {{else}}
+          {{if optionLabelPath (get option optionLabelPath) option}}
```
cibernox: can we save that if? At the end the user has to provide either a block or  the optionLabelPath. If the user provides none an assertion should warn him.

addon/templates/components/ember-power-select/single.hbs
```diff
@@ -15,7 +15,11 @@
     {{#if resultsLength}}
       {{#ember-power-select/options options=results highlighted=_highlighted selection=selection
         highlight=(action "highlight") select=(action "select") dropdown=dropdown as |option|}}
-        {{yield option}}
+        {{#if baseHasBlock}}
+          {{yield option}}
+        {{else}}
+          {{if optionLabelPath (get option optionLabelPath) option}}
```
cibernox: same

cibernox:
Added a couple comment. However I see one problem. 

The get helper is not available in ember 1.13. Therefore it concerns me that the tests are passing, since I thought that I added it to the test matrix on ember-try. I'll check it.

Do you want to add assertions in this PR or in a different one?

addon/templates/components/ember-power-select/multiple.hbs
miguelcobain: That is an inline if for calling `(get option optionLabelPath)` only if `optionLabelPath` and just render `option` otherwise.

The only way we could save that `if` is if `get` returned the entire object/param on a given pre initialized path. I.e, does `(get option "")` return `option`?

miguelcobain:
> Added a couple comment. However I see one problem.

Answered inline.

> The get helper is not available in ember 1.13. Therefore it concerns me that the tests are passing, since I thought that I added it to the test matrix on ember-try. I'll check it.

Some tests should definitely fail if there isn't a `get` helper.
Looking at Travis. It looks like ember 1.13 scenario isn't being tested.

> Do you want to add assertions in this PR or in a different one?

There's no hurry. I'll add them in this PR.

miguelcobain:
We might be stuck using ember-get-helper because of this: https://github.com/jmurphyau/ember-get-helper/issues/22

I don't want Ember 2.0 users getting deprecations.

cibernox:
In #42 I added 1.13 to the test matrix. Seems more broken that what I expected. I'll fix it today.

miguelcobain:
I've rebased and now Ember 1.13 tests fail, as expected.

If we're going to use the get helper, I guess we're blocked by https://github.com/jmurphyau/ember-get-helper/issues/22...

I don't see any reasonable alternative without adding a lot of code to the codebase.

cibernox:
Are you aware if you can load an addon conditionally?

miguelcobain:
I'm not aware, but that would certainly solve the problem! Will check that.

jmurphyau:
Should have the deprecations removed from ember-get-helper in v1.0.4 within the next day or so

cibernox:
Closed via #108

PR #43, opened 2015-10-15
add lato font

miguelcobain:
I was seeing an ugly arial font, until I found out why.


cibernox:
Let's look at it

miguelcobain:
The styles define "Lato" as the font family. However, if your pc doesn't have that font, displays Helvetica Neue, Helvetica or Arial.
Linux folks see Arial because Helvetica is proprietary.

cibernox:
I' so used to Helvetica that I thought that it was publicly available :open_mouth:

samselikoff:
yep source is here: https://github.com/samselikoff/ember-cli-mirage/tree/gh-pages

PR #45, opened 2015-10-15
Fix typos

btecu:
Fixed a few typos


cibernox:
Thanks you! I guess there is a lot more like those :(

Issue #49, opened 2015-10-16
Ember 2.0: Uncaught TypeError: _this2.matcher is not a function


srsgores:
I get this when I click on the dropdown, when passing in data with Ember Data 2.0:

> Uncaught TypeError: _this2.matcher is not a functionmatcher @ base.js:139filterOptions @ group-utils.js:96refreshResults @ base.js:142onOpen @ base.js:71closureAction @ ember.debug.js:21946open @ ember-basic-dropdown.js:91toggle @ ember-basic-dropdown.js:78toggle @ ember-basic-dropdown.js:58closureAction @ ember.debug.js:21946

srsgores:
I found the fix:

```
if (this.get('searchField')) {
      matcher = (option, text) => this.get("matcher")(get(option, this.get('searchField')), text);
    } else {
      matcher = (option, text) => this.get("matcher")(option, text);
    }
```

Notice I use ``get``, short for ``this.get()``, instead of trying to access the method directly from the class.

Want a PR?

cibernox:
Hey, good catch. I think this one slipped in a recent refactor. Please go ahead with the PR.

I'm worry tests didn't catch that one.

cibernox:
@srsgores Can you install the addon from master and check if it's fixed?

cibernox:
@srsgores Can you confirm this is not an issue anymore?

cibernox:
I'm closing this assuming that it's because you're on an version of Ember older that 1.13.9.

Let me know if it's not the case.

PR #50, opened 2015-10-20
Ensure matcher attr is initialized at correct time

opsb:
Was receiving an error about matcher not being a function at https://github.com/cibernox/ember-power-select/blob/master/addon/components/ember-power-select/base.js#L121. Adding matcher as a declared field fixed it.

### DIFF
--- a/addon/components/ember-power-select/base.js
+++ b/addon/components/ember-power-select/base.js
@@ -9,6 +9,25 @@ export default Ember.Component.extend({
   _loadingOptions: false,
   attributeBindings: ['dir'],
 
+  // Universal config
+  disabled: null,
+  placeholder: null,
+  loadingMessage: null,
+  noMatchesMessage: null,
+  selectedComponent: null,
+  dropdownPosition: null,
+  matcher: null,
+  searchField: null,
+  search: null,
+
+  // Select single config
+  searchEnabled: null,
+  searchMessage: null,
+  searchPlaceholder: null,
+  allowClear: null,
+
+  // Select multiple config
+
   // Lifecycle hooks
   didReceiveAttrs({ newAttrs: { options } }) {
     this.set('_loadingOptions', true);
@@ -116,9 +135,9 @@ export default Ember.Component.extend({
     const { _options: options, _searchText: searchText } = this.getProperties('_options', '_searchText');
     let matcher;
     if (this.get('searchField')) {
-      matcher = (option, text) => this.matcher(get(option, this.get('searchField')), text);
+      matcher = (option, text) => this.get('matcher')(get(option, this.get('searchField')), text);
     } else {
-      matcher = (option, text) => this.matcher(option, text);
+      matcher = (option, text) => this.get('matcher')(option, text);
     }
     this.set('results', filterOptions(options || [], searchText, matcher));
     this._resultsDirty = false;


miguelcobain:
Does this solve issue #49?

opsb:
I'm running against ember 1.13.7 so I'm not sure.

miguelcobain:
@srsgores - can you confirm that?

opsb:
Ok, I'll set a project up and take a look.

miguelcobain:
I was asking @srsgores, but it would be great if you could test that!

opsb:
@miguelcobain I did have a quick look but it doesn't look there's a version of ember-cli setup for v2 out of the box (don't have time to dig into that).

opsb:
Hopefully @srsgores can confirm.

miguelcobain:
Changing bower.json `ember` version is all that is needed.

cibernox:
I think that the proper fix would be to change `this.matcher` to `this.get('matcher')`.

I've been on holidays. Can you add a failing test case that the fix solves to be sure we never break this thing again?

opsb:
@cibernox I did try that fix first but it didn't work (it's still undefined). The most general fix is probably to do both. I've also discovered that the issue effects all attributes that are assigned in https://github.com/cibernox/ember-power-select/blob/master/addon/templates/components/ember-power-select.hbs. I'm going to update this PR to include those as well.

cibernox:
Can you paste the use case that breaks? When I did that refactor I relied in the test suite but clearly it's not caching something.

opsb:
It's just a vanilla

    {{#ember-power-select options=countries as |country|}}
      {{country}}
    {{/ember-power-select}}

It originally wouldn't render because of the matcher issue. Having fixed that I then noticed the search box wasn't appearing, same issue.

cibernox:
I'll check it now

2015-10-20 16:39 GMT+01:00 Oliver Searle-Barnes <notifications@github.com>:

opsb:
Just updated the PR. The attr declarations are duplicated though, probably worth extracting somehow.

cibernox:
Ok, i've been able to reproduce the bug. Actually is because you're in ember 1.13.7. 

I've faced this before. There is a bug that was solved in ember 1.13.9 in the new lifecycle hooks (didReceiveAttr, didInitAttrs, ect...) where if you don't call `this._super(...arguments)` from within those hooks the default behaviour is not triggered.

This is the commit where was fixed in ember itself: https://github.com/emberjs/ember.js/pull/12138

Theoretically this component is only compatible with ember 1.13.10+, but this is a very small change so I'll do it anymore. Hold tight.

cibernox:
This commit should fix it: https://github.com/cibernox/ember-power-select/pull/51

cibernox:
Can you try installing the component from master? I think that this should be fixed along with #49

opsb:
Giving it a go...

opsb:
Yep, master is working, nice work!

cibernox:
Also i've just fixed another unrelated bug, so pull again if you can :sweat_smile:

opsb:
Good stuff, ta!

srsgores:
Maybe I'm misinformed here, but shouldn't you guys just be using [ember try](https://github.com/kategengler/ember-try) for this?

miguelcobain:
We're using ember try.
Ember try just enables us to run tests in multiple scenarios.

Of course, if a certain bug isn't covered by the tests suite, ember try won't add anything. :smiley:

cibernox:
We do, but this addons is tested against the latest version of ember 1.13.10 and this bug is present for people using ember 1.13.8 and

cibernox:
We do, but this addons is tested against the latest version of ember 1.13.10 and this bug is present for people using ember 1.13.8 and

miguelcobain:
Ah, didn't realize that. :+1:

PR #54, opened 2015-10-22
Make `onchange` mandatory.

cibernox:
Since a select with no associate action is useless this takes the component
one step further towards explicitness.

Now the component won't do ANYTHING AT ALL when an element is selected, it will relay
completely in the `onchange` action defined by the user. The user has to define that
action and update the `selected` value to propagate the change to the component again.

An assertion has been added to lead the user in the good direction

### DIFF
--- a/tests/integration/components/ember-power-select-test.js
+++ b/tests/integration/components/ember-power-select-test.js
@@ -97,7 +97,7 @@ test('Click in the trigger of a closed select opens the dropdown', function(asse
 
   this.numbers = numbers;
   this.render(hbs`
-    {{#ember-power-select options=numbers as |option|}}
+    {{#ember-power-select options=numbers onchange=(action (mut foo)) as |option|}}
       {{option}}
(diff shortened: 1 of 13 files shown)


tests/integration/components/ember-power-select-test.js
```diff
@@ -1821,4 +1835,30 @@ test('the list of options can be customized using optionsComponent', function(as
...
+  this.numbers = numbers;
+
+  try {
+    this.render(hbs`
+      {{#ember-power-select options=countries selected=selected as |opt|}}
+        {{opt}}
+      {{/ember-power-select}}
+    `);
+  } catch(e) {
+    assert.ok(/requires an `onchange` function/.test(e.message));
```
miguelcobain: Qunit's `throws` function could be used in this test.
https://api.qunitjs.com/throws/

Not crucial.

cibernox: Done :)

PR #71, opened 2015-10-28
Trigger native events in tests instead of jQuery

cibernox:
Apparently jQuery events are different in many ways than native events, so
this is closer to the actual behavior of the addon in the wild.

Particularly jQuery events don't have a `.defaultPrevented` property to check,
and some upcoming functionalities require this.

Also update the dropdown addon to 0.3.2

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,7 +63,7 @@
   "dependencies": {
     "ember-cli-babel": "^5.1.3",
     "ember-cli-htmlbars": "0.7.9",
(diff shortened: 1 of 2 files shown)


miguelcobain:
If this kind of needs reappear a lot in the future, we should consider using somthing like syn library: https://github.com/bitovi/syn

cibernox:
Good to know that library.

I just copy-pasted this from stackoverflow tho, not a big deal.

PR #73, opened 2015-10-29
Expose power-select component as 'ember' seems redundant

achambers:


cibernox:
Actually makes a lot of sense. The only but is that you need to di a find and replace of `{{#ember-power-select}}` in the docs too.

cibernox:
And probably also deprecate the old name but keep it working for a while.

achambers:
Okie doke...Will do.  Any thoughts on how to deprecate something like that?

I'll do it for both repos.

cibernox:
```
import EmberPowerSelectComponent from 'ember-power-select/components/ember-power-select';
export default EmberPowerSelectComponent.extend({
  init() {
    this._super(...arguments);
    Ember.deprecate('Use new name');
  }
});
```

Something like that?

achambers:
Sweet

achambers:
@cibernox Any thoughts on what should go in the `options` hash for this stuff?

http://emberjs.com/deprecations/v2.x/#toc_ember-debug-function-options

cibernox:
It's optional, correct? If it's not put that it will removed in 0.4 and create a any id that makes sense. If it gives you troubles use `console.warn` and continue with your life, it's not a big deal.

cibernox:
I've applied the same change to ember-basic-dropdown. 

If "something happens" (:wink:) let me know and I continue from here.

cibernox:
I supossed that you are very busy, so I did it yesterday.

Closed vía #88

achambers:
I figured that might be the case. Sorry mate....Have been under the pump.
Cheers ;)

PR #76, opened 2015-10-30
increase z index of dropdown content

willrax:
as discussed in slack, this seems like it could be a sane default.

### DIFF
--- a/vendor/ember-power-select.css
+++ b/vendor/ember-power-select.css
@@ -7,7 +7,7 @@
 .ember-basic-dropdown-content {
   position: absolute;
   width: 100%;
-  z-index: 2;
+  z-index: 99;
   background-color: #FFF;
   width: auto; }
 


cibernox:
Woot!

I'll cut a new release over the weekend.

willrax:
Thanks! :tada:

miguelcobain:
Well, bootstrap modals have a default of z-index 1050: https://github.com/twbs/bootstrap-sass/blob/master/assets/stylesheets/bootstrap/_variables.scss#L283

:stuck_out_tongue_winking_eye:

cibernox:
@miguelcobain I wonder why that specific number. Seems too specific to be arbitrary.

cibernox:
Also seems that dropdowns have 1000.

willrax:
i just checked our css and our side panel is something like 900 :/

miguelcobain:
I think it really is arbitrary. AFAIK z-index values aren't relevant, just how they relatively compare to each other.

Perhaps they needed to stack other elements of the framework in depth. Modals, being the topmost of them (I think) ended up with that value. Don't really know.

jamesdixon:
Definitely arbitrary. I just ran into the same issue using [liquid-tether](https://github.com/pzuraq/liquid-tether) for modals and they too use 1050 for their z-index. Think it might be worth bumping up the default a bit more? Maybe to 1,475,348? :+1:

PR #77, opened 2015-10-30
change the z index on the correct class

willrax:
i just noticed that the class i changed was incorrect. I've put 999 here as this is what worked for us. I have no idea what a sane default for this value is.


cibernox:
Neither do I, so let's try with this and see how it goes.

willrax:
sounds good.

miguelcobain:
Like I've said, bootstrap's modals default z-index is 1050. Since this framework is so popular, I suggesting defaulting it to a larger value than 1050 to avoid possible future issues.

PR #79, opened 2015-10-30
ensure incoming text is a string

willrax:
if the text value that the user is searching for is an integer we need to ensure that its a string before running replacing on it.

### DIFF
--- a/addon/utils/group-utils.js
+++ b/addon/utils/group-utils.js
@@ -929,7 +929,7 @@ export function stripDiacritics(text) {
     return DIACRITICS[a] || a;
   }
 
-  return text.replace(/[^\u0000-\u007E]/g, match);
+  return `${text}`.replace(/[^\u0000-\u007E]/g, match);
 }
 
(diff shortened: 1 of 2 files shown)


addon/utils/group-utils.js
```diff
@@ -929,9 +929,9 @@ export function stripDiacritics(text) {
     return DIACRITICS[a] || a;
   }
 
-  return text.replace(/[^\u0000-\u007E]/g, match);
+  return `${text}`.replace(/[^\u0000-\u007E]/g, match);
```
cibernox: Out of curiosity, which one is faster, this or `String(foo)`?

willrax: No idea. I can do a bit of research. I’ve seen this technique in embers query parameters. I’ll get back to you.

cibernox: I answer myself: http://jsperf.com/number-to-string/2

willrax: Whoa, great website.

PR #85, opened 2015-10-31
Create components to encapsulate selected item logic

cibernox:
{{ember-power-select/single/selected}} and {{ember-power-select/multiple/selected}}

They can be replaced by the user


miguelcobain:
Brilliant! :+1:

Issue #142, opened 2015-11-20
Updating an array proxy through results


snewcomer:
Hi!  I'm looking to pass in an array proxy to `options` (that is populated at some point later from an xhr) and traced down the problem to breaking the cache on `results.[]` instead of just `results`.  Making this change works in production. Wrote a quick test but haven't got it to work yet; guessing I can't get that array proxy to update after its passed in for some reason.  Is your intended use to have the options be always wrapped in a promise (as your other tests show) or do you want to support something like this?  (noticed you mentioned in another post that if one wanted the loading message, all options must be wrapped in a promise)

    //line 41 in power-select/base.js
    resultsLength: computed('results.[]', function() {
      return countOptions(this.get('results'));
    }),

    test('xxxx wat', function(assert) {
      let done = assert.async();

      let numbers_0 = Ember.A([]);
      this.numbersArrayProxy = Ember.ArrayProxy.extend({
          content: Ember.computed('source.[]', function() {
              return this.get('source');
          })
      }).create({
          source: numbers_0 
      });

      Ember.run.later(function() { 
          numbers_0.pushObjects(numbers);
      }, 100);

      this.render(hbs`
        {{#power-select options=this.numbersArrayProxy onchange=(action (mut foo)) as |option|}}
          {{option}}
        {{/power-select}}
      `);

      Ember.run(() => this.$('.ember-power-select-trigger').click());

      assert.equal($('.ember-power-select-option').text().trim(), 'Loading options...', 'The loading message appears while the promise is pending');
      setTimeout(function() {
        assert.ok(!/Loading options/.test($('.ember-power-select-option').text()), 'The loading message is gone');
        assert.equal($('.ember-power-select-option').length, 20, 'The results appear when the array proxy is populated');
        done();
      }, 150);
    });

toranb:
@miguelcobain @cibernox @snewcomer care if I take this on over the weekend?

snewcomer:
@toranb yep go ahead!  I'll wait for your PR.

cibernox:
@toranb Please, be my guest. Although I foreseen that it won't be a very simple change.

I guess that you'll need to define `results` as a computed property that depends on the options and the searchText. Perhaps a PromiseArrayProxy?

toranb:
@cibernox dude it wasn't bad at all :)

https://github.com/cibernox/ember-power-select/pull/143

toranb:
@snewcomer you can close this issue out - thanks for the help !!!

PR #143, opened 2015-11-22
[BUG]: options did not refresh when empty array proxy was updated

toranb:
I was able to get a solid test to show the production bug Scott and I found recently ( mentioned in detail using issue #142 )

I could use a little feedback on "where this fits" in the integration test file (to make sure you are happy with it/etc). I also updated the doc string you had above the block (again -unsure if I did this as you expected).

Let me know what I can do to pull this in/ get a beta 4 some time this week :)

### DIFF
--- a/addon/components/power-select/base.js
+++ b/addon/components/power-select/base.js
@@ -38,7 +38,7 @@ export default Ember.Component.extend({
     return this.get('_searchText.length') === 0 && !!this.get('search') && !!this.get('searchMessage');
   }),
 
-  resultsLength: computed('results', function() {
+  resultsLength: computed('results.[]', function() {
     return countOptions(this.get('results'));
   }),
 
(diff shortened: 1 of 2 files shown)


cibernox:
That was much simpler than I thought. 

About updating the comments, I'm sure they are not completely up to date anyway. Before implementing the test suite I spent a couple hours gathering all the possible scenarios to tests and wrote comments for them before implementing them. Once they are implemented they are not really necessary.

Can you only add another test to check that it also works when updating pushing to an array that is not empty?

I'll push a new beta tomorrow.

Thanks! :bow:

toranb:
Sounds good! I went ahead and removed the comments I made at the top (not needed as you mentioned). I added the test you asked about (maybe it's worth mentioning -it didn't require any production code beyond what I added for the first -a valid regression test if nothing else).

If you need anything else let me know :)

cibernox:
Nope, this is perfect.

Thanks again!

PR #153, opened 2015-11-27
footer feature for ember-power-select

fadeojo:
Added footer functionality for ember-power-select and integration test for new feature.

### DIFF
--- a/addon/templates/components/power-select.hbs
+++ b/addon/templates/components/power-select.hbs
@@ -24,8 +24,10 @@
     class=class
     dropdownClass=dropdownClass
     extra=extra
-    as |option|}}
-    {{yield option}}
+    as |option footer|}}
+  {{#if (or (not-eq option 'noop') hasFooter)}}
+    {{yield option footer}}
+  {{/if}}
   {{else}}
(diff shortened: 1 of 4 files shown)


cibernox:
Hi. 

I appreciate the PR but precisely this kind of pattern where your yielded block can have several roles depending on a given argument you check against on each iteration is something I don't want to include in the default behaviour.

Hopefully soon Ember itself will support multiple blocks (there is an RFC about that) and then I will consider to augment the default behaviour with optional blocks like header & footer.

In the meanwhile my suggestion is that you create a custom `optionsComponent` with this extra feature

```hbs
{{#power-select options=a selected=b onchange=(action (mut b) optionsComponent="optionsWithFooter" as |opt|}}
  {{opt}}
{{/power-select}}
```

Your customized options component can extend directly from the default one and just have a different template where you apply this trick.

I'm sorry for rejecting this but I'd prefer this to be released as a separated addon built on top of ember-power-select. There is an addon section in the page, so let me know if you do it and I'll add it to that section.

Once ember has a less awkward way of having "named' blocks this functionality could make it into the core component.

Thanks for the PR.

PR #165, opened 2015-12-03
Refactor results and highlighted to be a computed property

cibernox:
The code is significantly simpler, and also shorter. I haven't measured
but I also hope that being lazy there is some other advantage.

This might be slightly incompatible since the selectedComponent doesn't
receive the hasPendingPromises argument, but the `results` argument is now
a PromiseProxy, so it can check the `isFulfilled` there.


miguelcobain:
These was one of my initial ideas for using cps for results. Big :+1:

cibernox:
Indeed, much simpler. And being lazy is aways a better approach.

PR #174, opened 2015-12-07
Divide test suite in several files

cibernox:
Rebuild times were very affected by the 2.5K lines of the file.
Now is 8x faster.


mszoernyi:
Great! Thanks

PR #176, opened 2015-12-08
Onfocus action

### LINKED ISSUE #172: Option to open search and set focus to it when component gets focus.
When entering the component from another form field the focus is set but it is not possible to immediately start typing and perform search. 
In a conversation with cibernox a proposed solution is to give the user an onfocus property to bind an action to. The action receives (select, event) and by calling select.actions.open() the search input can be automatically opened onfocus.

cibernox:
Example: Open the select when focused.

```hbs
{{#power-select options=options selected=foo onchange=(action (mut foo)) onfocus=(action "didFocus") as |option|}}
     {{option}}
{{/power-select}}
```

```js
actions: {
  didFocus(select /*, e */) {
    select.actions.open();
  }
}
```

Closes #172 

This also updates `ember-basic-dropdown` to 0.7.0-beta.1, which changes the strategy to open the select from `onclick` to `onmousedown`. This was necessary for this feature but is also aligned with the way real selects works and makes the component feel much snappy :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Master
 
+- [BREAKING] Update to ember-basic-dropdown 0.7.0-beta.1. This means that the component is opened/
+  closed using mousedown instead of click. This makes the component feel more snappy. It is unlikeliy
+  that this breaks real world usage but might break integration tests of people where people rely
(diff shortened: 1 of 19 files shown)


sweco-semtne:
Great work! And a big thank you.

PR #177, opened 2015-12-09
let options component render the parent element

miguelcobain:
For the loading, no matches and search messages I just wrapped them in the previous `<ul` as it seems the most straightforward way to reuse styles.

Also, I'm setting `class="ember-power-select-options"` from the outer template. This helps to not break tests. I don't think it makes any difference to the passed in component.


cibernox:
Just two things:

* Add an entry to the changelog
* Check that some custom examples, like the animated select work

PR #184, opened 2015-12-10
use triggerClass

miguelcobain:


addon/templates/components/power-select/multiple.hbs
```diff
@@ -1,5 +1,5 @@
 {{#basic-dropdown class=(readonly concatenatedClasses) dir=dir tabindex=(readonly tabindex) renderInPlace=renderInPlace matchTriggerWidth=true
-  disabled=disabled dropdownPosition=dropdownPosition triggerClass=(readonly triggerClass) dropdownClass=concatenatedDropdownClasses
+  disabled=disabled dropdownPosition=dropdownPosition triggerClass=concatenatedTriggerClasses dropdownClass=concatenatedDropdownClasses
```
cibernox: readonly

addon/templates/components/power-select/single.hbs
```diff
@@ -1,5 +1,5 @@
 {{#basic-dropdown class=(readonly concatenatedClasses) dir=dir tabindex=(readonly tabindex) renderInPlace=renderInPlace matchTriggerWidth=true
-  disabled=disabled dropdownPosition=dropdownPosition triggerClass="ember-power-select-trigger" dropdownClass=concatenatedDropdownClasses
+  disabled=disabled dropdownPosition=dropdownPosition triggerClass=concatenatedTriggerClasses dropdownClass=concatenatedDropdownClasses
```
cibernox: readonly

miguelcobain:
Couldn't add `readonly` on `opened` property. You have specific tests to assure it is mutable.
It is out of the scope of this PR to deal with that, IMO.

cibernox:
Yes, `opened` is the only exception. It's a mutable property by design. I also commented that in the API reference.

Issue #72, opened 2015-10-28
Add Custom Option


cottrellio:
Is there currently support for adding a custom option similar to the `create-item` action on `ember-cli-selectize`?

miguelcobain:
This is also a use case I will need.
Basically, tag support, @cibernox.

cibernox:
It will land, probably next week, but probably not like an magic config option but as an action you can implement :)

miguelcobain:
Sure.
But we need to display some text in the dropdown, don't we?
Like "Create {{text}} option".

cottrellio:
Yeah that would be nice.

catataw:
+1

felixkiss:
> It will land, probably next week

@cibernox What is the timeline on this feature? I couldn't find it in the CHANGELOG up to `0.6.0-beta.4`, so I assume it was delayed. What are the reasons?

cibernox:
@felixkiss This is the PR: https://github.com/cibernox/ember-power-select/pull/150

Basically I added on `onkeydown` action that you can use to do whatever you feel like. The most obvious use case is detect that the user pressed enter, take the text of the event's target and create a new option with it. Typical select that allows to create new tags on the fly.

I'm sitting on the fence about that is the best way of preventing the default behaviour (close the select on enter by example). Will be merged today.

cibernox:
Closed via #150

felixkiss:
@cibernox 

I would like to implement behavior as seen in [ember-cli-selectize](https://github.com/miguelcobain/ember-cli-selectize):
![screen shot 2015-11-29 at 22 07 27](https://cloud.githubusercontent.com/assets/256075/11460093/2e969a32-96e6-11e5-8610-dbd395ebfe92.png)

- Show option `Add <current search term>...` above search results

Can you give me a hint on how to accomplish this? I think I would need the following:

- Ability to inject a template for an additonal search result item (in this template, I need to have access to at least the current search term, to not show `Add ...` on empty search term)
- Ability to select this custom search result item as any search result item (up/down arrow keys work; enter = selection)
- Upon selecting this custom search result item, an action should be fired
- Visual distinction between existing search result items and this custom create item would be really nice to have. Selectize really does a great job on this.

cibernox:
Sure, I think that did a similar example somewhere. Let me search...

cibernox:
[...]

PR #185, opened 2015-12-10
Add action to create custom option

lan0:
This is an attempt to allow a public API for creating an option: `{{power-select oncreate=(action "itemCreated") ...}}`. The goal is to have a highlightable (up/down arrows + enter) option that fires the action (`Add xyz...`).

There is no support for keyboard navigation, seems like it would be hard to implement? Also it doesn't work for `multiple=true` (`Uncaught TypeError: Object.defineProperty called on non-object` - no idea why).

The solution in https://github.com/cibernox/ember-power-select/issues/72#issuecomment-160479966 does not work in a general case because for creating a new (fake) option the component needs to know which attributes are displayed in the option template. Using a custom `optionsComponent` that deals with item creation falls short because the `optionsComponent` can't communicate an action to the code using the `power-select`, informing it that an item should be created.

Do you want to support such an `oncreate` action in this addon? It would be very nice to have this kind of functionality ready to use in different parts of the application over a simple API.

### DIFF
--- a/addon/templates/components/power-select/multiple.hbs
+++ b/addon/templates/components/power-select/multiple.hbs
@@ -12,7 +12,9 @@
       removeOption=(action "removeOption" dropdown)
       handleKeydown=(action "handleKeydown" dropdown)
     )) as |select|}}
-
+      {{#if createEnabled}}
+        <ul class="ember-power-select-options "><li {{action "create" dropdown searchText}} class="ember-power-select-option">Add {{searchText}}...</li></ul>
+      {{/if}}
       {{#if showLoadingMessage}}
         <ul class="ember-power-select-options"><li class="ember-power-select-option">{{loadingMessage}}</li></ul>
       {{/if}}
--- a/addon/components/power-select/base.js
+++ b/addon/components/power-select/base.js
@@ -60,6 +60,10 @@ export default Ember.Component.extend({
     return countOptions(this.get('results'));
   }),
 
+  createEnabled: computed('oncreate', 'searchText', function() {
+    return this.get('oncreate') && ! Ember.isEmpty(this.get('searchText'));
+  }),
+
   // Actions
   actions: {
     open(dropdown, e) {
@@ -76,6 +80,10 @@ export default Ember.Component.extend({
 
     search(dropdown, term /*, e */) {
       this._doSearch(term);
+    },
+
+    create(dropdown, term) {
+      this.get('oncreate')(term, this.buildPublicAPI(dropdown));
     },
 
     handleKeydown(dropdown, e) {
--- a/addon/templates/components/power-select.hbs
+++ b/addon/templates/components/power-select.hbs
@@ -27,6 +27,7 @@
     class=class
     dropdownClass=dropdownClass
     extra=extra
+    oncreate=oncreate
     as |option term|}}
     {{yield option term}}
   {{else}}
--- a/addon/templates/components/power-select/single.hbs
+++ b/addon/templates/components/power-select/single.hbs
@@ -12,14 +12,16 @@
       clear=(if allowClear (action "select" dropdown null))
       handleKeydown=(action "handleKeydown" dropdown)
     )) as |select|}}
-    {{#if searchEnabled}}
-      <div class="ember-power-select-search">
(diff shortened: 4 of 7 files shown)


addon/templates/components/power-select/multiple.hbs
```diff
@@ -12,7 +12,9 @@
       removeOption=(action "removeOption" dropdown)
       handleKeydown=(action "handleKeydown" dropdown)
     )) as |select|}}
-
+      {{#if createEnabled}}
+        <ul class="ember-power-select-options "><li {{action "create" dropdown searchText}} class="ember-power-select-option">Add {{searchText}}...</li></ul>
+      {{/if}}
```
miguelcobain: Please make the message a computed property, like the other messages. One might want to customize this (e.g non-english users).

cibernox:
I think that this is already possible, and I have an example somewhere. Let me check.

cibernox:
Yes, there is an exact example like this on the docs: http://www.ember-power-select.com/docs/action-handling

Is this a viable option? I prefer to have a low-level API like `onkeydown` because some people might want to create options on Enter, and other people might want to create options on TAB. 

If this doesn't solve your use case I'd love to explore how to improve the component to give more power to the user.

cibernox:
After reading the PR a second time I think that I understand your problem better. You want this "fake" option to be outside of the of the options list.

Yes, that is not straightforward right now, but might be possible since #177. I'll check it.

miguelcobain:
With #177 users could provide a tagless optionsComponent and then render two lists there (or a <p and a list or whatever is needed).

felixkiss:
I want to summarize what @lan0 and I had in mind, when we talked about this today. Hopefully it is useful to everyone trying to think about a possible implementation.

# Current state
 - It is possible to completely customize `power-select` by handling a combination of `search`, `onchange`, and `onkeypress`
 - Logic for all this is offloaded to the user of the `power-select` instance (e.g. a component or controller)
 - During the time the user interacts with the `power-select`, the options the user passes into `power-select` could be populated with a temporary fake option. This can be potentially harmful, if it is an ember-data relationship and it could be accidentally persisted if the related model is saved during this time.

# New model
 - [x] It is completely optional. You "activate" the create-item functionality by passing an action into `power-select` to handle the creation process of a new option. This obviously needs to be handled by the user, because the data could come from anywhere (e.g. ember-data, some external API, etc.). If you don't pass anything in, nothing changes. This is backwards compatible. (Possible as of bcade91)
 - [x] If you type into the search box and create-item is activated, there is one option to select/highlight above all results `Add "<search term>" ...`. (Possible as of bcade91)
 - [ ] The user should be able to navigate by moving the mouse cursor above the option or by using the arrow keys. (It looks like a rather invasive change to the current code e.g. in `handleVerticalArrowKey()`)
 - [x] The list of options that are passed into `power-select` only change when the user selects the `Add "Foo Bar" ...` item above the list. There is no temporary, fake thing that could have side effects on your app.
 - [ ] Optionally, the user can pass in a custom message, as suggested by @miguelcobain for i18n or other reasons. If not, `Add "<search term>" ...` (or something similar) is used.
 - [x] Optionally, the user still has the ability to get other behavior by handling `search, `onchange`, or `onkeypress`. (This is possible in master, and should continue to work)
 - [ ] It should work with `multiple=true` as well (as mentioned by @lan0, this doesn't work as of right now)
 - [ ] All functionality should be covered by tests
 - [ ] All functionality should be documented in the docs

This is intended to be a one-size-fits-many approach. If you like it, there is almost no additional work to be done. If you don't, you can ignore it and roll your own solution to the problem. I think this would be useful to a lot of users.

Since we don't know the codebase very well, maybe the two of you who are familiar with it could implement the remaining things or point us in the right direction?

Let me know what you think about this.

cibernox:
It's not that I disagree, is that I think that is would be easy to do with the existing primitives and in a simper way, without having to control keystrokes yourself.

There is more edge cases in arrow navigation of what it seems and by having a fake option you avoid them. I'll come with an example today.

cibernox:
Closed since https://github.com/cibernox/ember-power-select-with-create is a thing now

PR #192, opened 2015-12-11
Add ARIA roles

Dremora:
Add ARIA roles to dropdown menus.

In case you're wondering, in nested menus, `<li>` containing child submenu is still a `menuitem`: https://www.w3.org/WAI/GL/wiki/Using_ARIA_menus#Example_Code


cibernox:
I'm not entirely convinced `menu` and `menuitem` are the most correct values. I'm not an ARIA guru but I've seen other selects using `listbox` and `option` (and `combobox` for multiple-select)

Dremora:
Good point. Looks like `listbox` and `option` is indeed a better fit. In regards to `combobox`, it should be used for the `<input>`.

PR #194, opened 2015-12-12
Expose test helpers to users of the package

felixkiss:
This enables meta-addons like `ember-power-select-with-create` or projects using
`ember-power-select` to use the test helpers in their own tests.

Usage:
```
import typeInSearch from '../../helpers/type-in-search';

Ember.run(() => typeInSearch('Foo Bar'));
```

Reference:
 - http://williamsbdev.com/posts/Ember-CLI-test-support/
 - http://ember-cli.com/extending/#addon-scenarios


cibernox:
I'm thinking that it could be nice to expose a `selectChoose` and `selectSearch` acceptance helper that abstracts interaction with the component completely.

The end user types
```js
selectSearch('.path-to-power-select', 'new');
selectChoose('.path-to-power-select', 'New york');
```

felixkiss:
Good idea. Either what you had in mind or something like `ember-pikaday` has.

https://github.com/edgycircle/ember-pikaday#test-helpers

PR #196, opened 2015-12-13
Put the test helpers in /test-support in a file with the name of the addon

cibernox:
@felixkiss This will make you amend your work in ember-power-select-with-create

I'm moving all helpers to a single file named after the addon. I copied this strategy from ember-simple-auth.

That way is easier for end users to consume them, because instead of
```js
import typeInSearch from '<app-name>/tests/helpers/type-in-search';
```

they will do

```js
import { typeInSearch } from '<app-name>/tests/helpers/ember-power-select';
```

This is easier to remember and is hard to have names collisions if helpers
have common names like `select` or `choose`.


felixkiss:
:+1:

PR #225, opened 2016-01-04
add sass variable for the class multiple options

tyleriguchi:
Add a sass variable to change the border for elements that have the multiple option class.


cibernox:
Tests are red in master since I updated ember-cli. Let me fix that. Other than that looks good :+1:

cibernox:
Apparently the cause is a bug in ember-cli (https://github.com/ember-cli/ember-cli/issues/5300) that is already being changes, so I'll merge this anyway.


Thanks!

tyleriguchi:
Great thanks for the quick merge! :+1:

PR #227, opened 2016-01-05
Few more sass variables and a typo fix

tyleriguchi:
Added a couple more sass variables to handle padding for the trigger classes and the multiple option class. Also fixed a small typo.


cibernox:
Also very reasonable :+1:

Issue #249, opened 2016-01-21
Don't re-compute results based on the search term changing


chrisgame:
From a performance point of view it would be good to not always throw away the search results just because the search term has changed.

miguelcobain:
@chrisgame your search function could implement that.
I doubt that a caching feature is desired in power-select core.
Maybe a power-select addon.

cibernox:
@miguelcobain Actually I've tracked some performance problems with this with @chrisgame 

The problem is that with async searches, you change the `searchTerm` and the entire `results` computed property is staled. Then you perform an async search and since it's async, you return the existing results while waiting for the promise to resolve. At that point you trigger a re-render even if the result haven't changed yet, because the `searchTerm` did, and ~X00 ms later, when the promise resolves, you re-render again.

With long collections with non-trivial html the select becomes slow because you re-render once because the search term changed (let's say, 300ms) and immediately after you re-render again (another 300ms).

It's true that is a pathological use case, but perhaps this needs stop making results a computed property, even if it was very convenient. 
Also, since `Ember.ArrayProxy` is being deprecated (https://github.com/emberjs/ember.js/pull/12843) this approach is screwed anyway :(

cah-briangantzler:
Ember data depends on PromiseArray with is basically an array proxy with promise. They may be deprecating it in ember but leaving it in ember data. Personally I think PromiseObject and PromiseArray should be in ember NOT in ember data as I have used them outside of ember data.

So if they deprecate it, you should be able to created it again but using the same mixins that ember data uses. This unfortunately puts a dependency on ember data (hence why I think those entites should be in ember)

I seen your questions on the above thread but they had not yet answered. Interested in the answer

cibernox:
`DS.PromiseArray` is just `Ember.ArrayProxy(Ember.PromiseProxyMixin)`, so you really don't need Ember-data, just this line. That said, I've already refactored in a branch the component to not use `ArrayProxy`, and the code is not bad either. 
I also want to check if we have some performance gain from that, but I still have 1 failing test I need to check.

PR #250, opened 2016-01-21
Stop using array proxy

cibernox:
Since `ArrayProxy` is being deprecated (probably) this stops using it. This also helps tackle some unnecessary re-renders discovered and reported in #249

### DIFF
--- a/addon/components/power-select-multiple/trigger.js
+++ b/addon/components/power-select-multiple/trigger.js
@@ -69,6 +69,7 @@ export default Ember.Component.extend({
 
   // Methods
   handleClose() {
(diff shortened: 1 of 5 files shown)


cibernox:
After some benchmarking, opening the dropdown is more or less the same (5/10% faster, but so small that could be fruit of chance) but filtering is exactly 2x faster (because the code in master triggers render twice while this one one once)

cibernox:
Closes #249

cibernox:
@miguelcobain Check out the code. It's a bit longer but given how complex the computed property ended being, I'd even say that this one is easier to follow.

PR #253, opened 2016-01-22
Fix functions in public API

### LINKED ISSUE #252: can't trigger a async search onOpen
I'm trying to cause the search action to be invoked when the dropdown is opened. Basically, I want to populate the dropdown with options before the user has to type anything. Here's my javascript:

        actions: {
    onSelectOpen: function(select, e) {
      console.log('open');
      console.log(select);
      select.actions.search('');
    },
    
    onChange: function(choice) {
      this.set('testSelected', choice);
    },

    search: function(query) {
      return ['hi', 'bye'];
}

and my hbs:

{{#power-select-multiple
	   onopen=(action "onSelectOpen")
	   search=(action "search")
	   selected=rule.value
	   renderInPlace=true
	   disabled=disableValueEdit
	   onchange=(action "onChange") as |name|}}
	  {{name}}
	{{/power-select-multiple}}

when i click on the component, onSelectOpen is triggered, but 'search' is not.
I'm probably just missing something, but I'd appreciate some guidance.

cibernox:
Closes #252

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Master
 
+# 0.8.2
+- [BUGFIX] Update ember-basic-dropdown again to fix bad positioning issue.
+- [INTERNAL] Stop using ArrayProxy internally. That saves some double render, improving performance
+  when filtering greatly (+100%) and very slightly in initial openinig (+5%)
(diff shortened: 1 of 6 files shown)


cibernox:
Sass is failing randomly.

PR #263, opened 2016-01-27
Remove tabindex completely when the component is disabled

### LINKED ISSUE #258: Trigger should not focus on click when disabled=true
Right now the trigger will gain a focus ring when clicked on while `disabled=true`. 
This is not consistent with native form components.

cibernox:
Closes #258


martndemus:
:+1:

PR #280, opened 2016-02-01
Added a variable for line-height units

CAH-MaggieRoss:
Calculations using line-height to determine the height of elements had been hardcoded to use em units, creating an issue for applications not using a unitless value for line-height or not using em units for line-height. 

**Changes:**
Replaced the em units with a variable in ember-power-select.scss and defaulted it to em in the variables. Leaving the variable blank with quotes("") when overriding will allow no unit to be used for the calculations, eliminating redundant units that would create an invalid value if line-height already has a unit.

### DIFF
--- a/app/styles/ember-power-select.scss
+++ b/app/styles/ember-power-select.scss
@@ -19,7 +19,7 @@ $ember-basic-dropdown-content-background-color: $ember-power-select-background-c
   background-color: $ember-power-select-background-color;
   line-height: $ember-power-select-line-height;
   text-overflow: ellipsis;
-  min-height: #{$ember-power-select-line-height}em;
+  min-height: #{$ember-power-select-line-height};
   user-select: none;
   -webkit-user-select: none;
   color: $ember-power-select-text-color;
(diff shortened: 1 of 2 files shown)


cibernox:
@CAH-MaggieRoss I'm thinking that perhaps is simpler to just make `$ember-power-select-line-height` contain both the value and the unit.
That way whoever is wants to change the value can just do `$ember-power-select-line-height: 2rem;`.

You change is backwards compatible, but this will go in the upcoming 0.9 cycle, so we can afford a breaking change now if that makes the styles easier to use. 

What do you think?

CAH-MaggieRoss:
My initial thought for fixing this was much the same, to simply remove the em units and let people define the units in the line-height value, as I had something like `$ember-power-select-line-height: $line-height-normal;`, which translated to `$ember-power-select-line-height: 2.5vmax;`, and is what broke when you added the ems. I realised the problem with that, however, is if someone did want to have a unitless line-height value like `1.5;`, then your two height calculations on lines 22 and 158 in ember-power-select.sass will break as they won't get a unit at all--maybe a better name for the new variable would have been `$ember-power-select-height-unit-for-line-height` or `$ember-power-select-height-calc-unit` since the unit is only needed on the calculation if the user doesn't want to set a unit on line-height.

cibernox:
@CAH-MaggieRoss That was a good point. I was assuming that this value was unitless, but it doesn't has to be the case.

Now I've added a guard using the `unitless` helper, so I think that we can provide only one variable that can contain units.

cibernox:
Relevant commit: https://github.com/cibernox/ember-power-select/commit/ec2eb293a790f1a9dec09309700de102aa18e37e

CAH-MaggieRoss:
Perfect! Thank you!

PR #300, opened 2016-02-09
Show "No resuls" only when search was performed or no search callback provided

Dremora:


cibernox:
@Dremora I've added tests to your change in #302

PR #307, opened 2016-02-10
Empty search, when term is empty and options is promise

RuslanZavacky:
If options passed to the component, is a promise and search is enabled via `search=(action "search")` component do not return initial results, when term is empty.

```javascript
options: computed(function() {
  return Ember.RSVP.Promise.resolve(function() {
    return [{id: 1, text: 'demo'}, { id: 2, text: 'demo2' }];
  });
});
```

It will work on initial search, but if you remove search term, it will show empty results.

### DIFF
--- a/addon/components/power-select.js
+++ b/addon/components/power-select.js
@@ -298,7 +298,13 @@ export default Ember.Component.extend({
     if (isBlank(term)) {
       let options = this.get('options') || [];
       this.activeSearch = null;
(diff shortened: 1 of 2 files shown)


cibernox:
I've added a commented test case in master: https://github.com/cibernox/ember-power-select/blob/master/tests/integration/components/power-select/custom-search-test.js#L534-L554

Can you rebase, uncomment and verify that it works? Also, there is a jshint error.

PR #316, opened 2016-02-11
[FEATURE] Typing in focused selects highlight or select the first match as native selects do

cibernox:


cibernox:
Video of the new feature: http://quick.as/zDaQSXX2V

Typing on selects without searchbox, or those with seachbox that are closed behaves like native selects and select or highlight the first match. That search is reset after 1s of idle time.

So typing "tw" will select/highlight (or highlight if the select is open) the option `two`, but typing "tw" + (wait 1s) + "o" will end up selecting/highlighting `one`.

jeffstagedoor:
Thanks for that new feature!
Unfortunately this only works with a simple one-dimentional option-list.

I for example have this as source:
gender: [
    {"code": 0, "label": 'female'},
    {"code": 1, "label": 'male'}, 
    {"code": 2, "label": 'other'},
]

_with:_

{{#power-select 
	options=genderList
	searchEnabled=false
	searchField='label'
	selected=genderSelected
	onchange=(action "selectGender") 
	as |gender|}}

	{{gender.label}}
{{/power-select}}

which will not select anything when typing a matching letter.

Maybe it would be possible to specify the field to 'search' for like I did, even if the normal search is disabled?

cibernox:
That is exactly how it works. Are you in the latest version?

jeffstagedoor:
Yes, I updated yesterday, so I'm on 0.8.6
Couldn't get it to work though.
I will do further test tomorrow, maybe it only didn't match because of typing lowercase while having Camelcased source-data.

cibernox:
The feature was added in 0.9. Try that one.

PR #332, opened 2016-02-28
Fixes default focus outline

ktrhn:


cibernox:
I think that this is not achieving what I wanted to do. I don't want the outline to be `none` (That is up to you, I don't want to change the default of your browser out of the box).

I'm actually checking if the outline is defined before adding it to the styles: 
https://github.com/cibernox/ember-power-select/blob/f59dc7dd53d0af1722b6abce360b923c219a05ed/app/styles/ember-power-select.scss#L131-L133

Although I'm thinking that perhaps default to `initial` would be more straightforward and not require `@if/@else`

ktrhn:
Oh...I didn't look closely enough. Sorry. Initial does not seem to work on outline. So your current solution is correct.

PR #333, opened 2016-02-29
Seperated handleKeyDown into seperate overridable methods

cah-briangantzler:
I would like to implement a selectOnTab property to allow the option to be selected when tabbed. It appeared I had to override the entire handleKeyDown action, but this was problematic based on the action keydown, if I did my code first, the action wouldn't fire.

It would be easier to just override the key I wanted to handle. To do this I split out the handling of each key into a method that could be overridden in an extended class.

This pull request does nothing but split the handling into those methods.

Once accepted I will follow up with a PR for implementing the selectOnTab. I did not include it in this request as I think this request would be useful to anyone extending the class and gives you the opportunity to not to take the selectOnTab functionally while still giving me the ability to implement it.

### DIFF
--- a/addon/components/power-select.js
+++ b/addon/components/power-select.js
@@ -170,17 +170,13 @@ export default Ember.Component.extend({
       if (onkeydown) { onkeydown(this.buildPublicAPI(dropdown), e); }
       if (e.defaultPrevented) { return; }
       if (e.keyCode === 38 || e.keyCode === 40) { // Up & Down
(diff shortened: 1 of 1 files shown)


cibernox:
This looks cleaner. 

Wait a bit before implementing the other PR, because I think at some point I considered that and I can't remember why I didn't.

cibernox:
Thanks!

PR #338, opened 2016-03-02
_doSearch should handle options as a promise.

GCheung55:
`this.get('options')` could return a promise, so `_doSearch` is updated to handle it. Otherwise, `_doSearch` could be triggered and improperly excute `this.filter` on a promise, instead of a promises fulfilled value.

### DIFF
--- a/addon/components/power-select.js
+++ b/addon/components/power-select.js
@@ -337,7 +337,14 @@ export default Ember.Component.extend({
       if (searchAction) {
         this._performSearch(searchAction, term);
       } else {
-        this.setProperties({ results: this.filter(this.get('options'), term), searchText: term, lastSearchedText: term });
+        let options = this.get('options');
+        if (options.then) {
+          options.then((data) => {
+            this.setProperties({ results: this.filter(data, term), searchText: term, lastSearchedText: term });
+          });
+        } else {
+          this.setProperties({ results: this.filter(this.get('options'), term), searchText: term, lastSearchedText: term });
+        }
       }
     }
   },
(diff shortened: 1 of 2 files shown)


GCheung55:
I still need to write tests.

addon/components/power-select.js
```diff
@@ -337,7 +337,14 @@ export default Ember.Component.extend({
       if (searchAction) {
         this._performSearch(searchAction, term);
       } else {
-        this.setProperties({ results: this.filter(this.get('options'), term), searchText: term, lastSearchedText: term });
+        let options = this.get('options');
```
cibernox: Totally right. I actually did the same in lines 110~116, but not here. This PR just needs tests to prevent a regression, probably in the `general-behaviour-test.js` file.

addon/components/power-select.js
```diff
@@ -337,7 +337,14 @@ export default Ember.Component.extend({
...
+        if (options.then) {
+          options.then((data) => {
+            this.setProperties({ results: this.filter(data, term), searchText: term, lastSearchedText: term });
+          });
+        } else {
+          this.setProperties({ results: this.filter(this.get('options'), term), searchText: term, lastSearchedText: term });
```
cibernox: Use the options variable defined in line 340.

cibernox:
I will release a new version just after merge this.

GCheung55:
@cibernox Tests added and reused `options`.

cibernox:
Thanks. I'll release a new version in a moment

Issue #331, opened 2016-02-27
power-select-multiple selected items in RTL layout are not aligned correctly


behnoodk:
`power-select` renders correctly when `dir=rtl` is set. the only thing that has not been included in the default CSS is the alignment of the selected items (span tags) and input element in `power-select-multiple`. this can be resolved by adding these two classes.

```
.ember-power-select[dir=rtl] .ember-power-select-trigger .ember-power-select-multiple-option,
.ember-power-select[dir=rtl] .ember-power-select-trigger .ember-power-select-trigger-multiple-input{
  float: initial;
}
```

right now the default, for both `ltr `and `rtl `is `float:left`.
let me know what do you think, I can make a pull request.

cibernox:
I'm not using any rtl language at the moment, so I'm not surprised I missed stuff.

That PR sounds good :shipit:

PR #347, opened 2016-03-04
Fix alignment of selected items in power-select-multiple for RTL layout

behnoodk:
added css classes for `ember-power-select-multiple-option` and `ember-power-select-trigger-multiple-input` in RTL layout to align them to the right, as discussed in #331

### DIFF
--- a/app/styles/ember-power-select.scss
+++ b/app/styles/ember-power-select.scss
@@ -232,6 +232,12 @@ $ember-basic-dropdown-content-background-color: $ember-power-select-background-c
 .ember-power-select[dir=rtl] {
   .ember-power-select-trigger {
     padding: $ember-power-select-trigger-rtl-padding;
+    .ember-power-select-multiple-option {
+      float: initial;
+    }
+    .ember-power-select-trigger-multiple-input {
+      float: initial;
+    }
   }
   .ember-power-select-status-icon {
     left: 5px;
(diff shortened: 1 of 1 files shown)


cibernox:
I trust you, but could you explain how `float: initial;` fixes this? I'd expect you use `float: right;` when the component is in rtl mode, and I fail to see how this works.

behnoodk:
Since this is only applied to power-select components with `[dir=rtl]`, `float: initial;` is in fact the same as `float:right;`. I set it to `float:initial;` because there was no reason to set it to `float:right;` explicitly.

The reason we need these two classes in the first place, is that the `float `property of `.ember-power-select-multiple-option` and `.ember-power-select-trigger-multiple-input` is set to `float:left;` . Had they been set to `float:initial;` it would have worked for both `dir=rtl` and default (`dir=ltr`).

Adding these two classes seemed a simple solution, one which I could easily test and there is no risk of breaking anything else.

cibernox:
I just created this quick jsbin (http://jsbin.com/toguji/edit?html,css,output) to learn about that property when float, and I think that `float: initial` is like telling `float: none;`. Try it.

So apparently it's not equivalent to float right.

The reason why I made this to float instead of just be inline-block items is because I don't want that significant whitespace in the templates affects the the separation between items.

behnoodk:
You are right! I changed `float:initial;` to `float:right;` , so they always float, and to make sure whitespace in the templates won't affect the the separation between items.

cibernox:
Thanks!

behnoodk:
Thank you! for your great work on this addon.

PR #353, opened 2016-03-08
Enable babel polyfill

gregjopa:
This PR enables the babel polyfill through ember-cli-babel. This polyfill is needed to support several ES6 features listed here: https://github.com/babel/ember-cli-babel#features. Previously, the `for (let val of array)` statements in the tests/dummy/app were causing problems in IE 11 (ex: https://github.com/cibernox/ember-power-select/blob/master/tests/dummy/app/controllers/public-pages/cookbook.js#L30). With this fix the docs website works successfully in IE 11.


cibernox:
Good catch

cibernox:
Thx

gregjopa:
:+1:  Happy to help

PR #365, opened 2016-03-13
Wrap yielded selected item in span

cah-briangantzler:
All text in the trigger is wrapped in some element except the yielded selected item. To allow for styling, wrapped this text in a span.

Did not opt for wrapping in the actual yield when ember-power-select is used, because that yield may also be used in options where is it already wrapped in an [LI element](https://github.com/cibernox/ember-power-select/blob/master/addon%2Ftemplates%2Fcomponents%2Fpower-select%2Foptions.hbs#L23)


cibernox:
Makes total sense :+1:

cibernox:
Thanks!

Issue #364, opened 2016-03-13
internal search action may no longer be called


cah-briangantzler:
Although you give an example of debounce, this is in the parent (controller/component) and would have to be rewritten every where you want a debounce.

I have extended the component to add min char and debounce by overriding the internal search action. This was awesome because the search action called nothing but _doSearch and was a great override point.

In version 0.9.0 you have streamlined things to where I believe the internal search action is no longer being called. I believe this is also the case of highLight and select. The choose action is still preserved. You can see this in the [publicAPI](https://github.com/cibernox/ember-power-select/blob/master/addon/components/power-select.js#L144). Could we not remove those actions but instead bring them back by having the publicAPI call them similiar to choose.

cah-briangantzler:
Looking at the code, this issue it still applicable to the 10-beta.2

cibernox:
This is not public API but it's true that is more idiomatic. I'll check it today, finally.

cah-briangantzler:
Looked at this code, think this does what I wanted it to do. Thanks.
Looking forward to a release that includes it to try it out.

cibernox:
I think I already did one release yesterday including this.

cah-briangantzler:
Sorry, yes you did do a tag. I am use to looking for actual releases (they show up big on the releases page), and missed the collapsed tags at the top. Sorry

PR #367, opened 2016-03-14
[RFC] Add oninput action

cibernox:
This is a last call for comments (and first) about a feature added to give a little more freedom to users that want do things when the user types, other than just searching.

#### Motivation
As of today, any use that wants to perform any custom operation when the user interacts with the searchbox has to use either the `onkeydown` action (which is cumbersome and doesn't gets triggered with copy-paste events though the mouse) or use the `search` action, although the feature being implemented is not a search per se.

Although I want to keep the API minimal, I favour adding low-level APIs like `on***` to respond to events over context specific APIs like `onEscPressed`. This seems generic enough.

#### `oninput` description

This action will be the action called when the user uses types in the searchbox of a select (single or multiple).

##### Arguments

Current implementation:
The action receives 3 arguments `oninput(text, select, e)`
* `text` The value of the input
* `select` the public API of the component
* `e` the raw DOM event.

Possible alternatives:
* Don't pass the text, just the public API and the event, and let the user extract the value from it. I kind of like it. Less magical.

##### Behaviour
* No behaviour by default.
* If an `oninput` action is provided, that action will be executed before the search.
* If the action passed to the `oninput` return false, then the default behaviour (trigger a search) is not executed automatically (the user still can trigger it manually since the public API has a `search` action). **why `return false` and not `preventDefault()`? ** Because input events are not cancellable.

#### Example: Multiple select that tokenizes inputs (by space).
```hbs
{{#power-select-multiple options=opts selected=sel oninput=(action "tokenizeAndSearch") onchange=(action (mut sel)) as |opt|}}
  {{opt}}
{{/power-select-multiple}}
```

```js
actions: {
  tokenizeAndSearch(text, select, e) {
    let parts = text.trimLeft().split(/\s+/);
    let tokens = parts.slice(0, -1);
    let last = parts.slice(-1)[0];
    this.get('sel').pushObjects(tokens); // Create tags with the tokenized elements
    select.actions.search(last); // If there is remaining text it is used for the search.
    return false; // Do not perform the default behaviour
  }
}
```
Simple video of the behaviour attained with this little snippet.
http://quick.as/8ynwtoxn5

#### What if the trigger has more than one input, and not all of them are for searching?

a) The user can fire this event and distinguish the source by the `event.target`
b) Fire another action instead, pased using the `extra` option.
c) In ember 2.3+, use contextual component to pass a component that is already bound to a different action.

#### Breakages

The public API of the component is totally backwards compatible.
The semi-public API (the implicit API api that components passes to EPS must use) has changed, since now the searchboxes should call `handleInput` instead of triggering the search automatically.

@Dremora @cah-briangantzler @miguelcobain :+1: ? :-1: ? Caveats?


cah-briangantzler:
I have always been a fan of returning true to continue rather than false to stop. The have intercepted the action, they can return true to continue. Returning nothing will return undefined, a falsy value, yet still continue. Actions bubble by default unless you intercept, in which case you must return true to continue the bubble.

As noted in [this issue](https://github.com/cibernox/ember-power-select/issues/364) search action was no longer being called. You have the default action (line 181) calling the action search (which I want), but if the user decided to call the publicAPI search, they would bypass the action and call _doSearch directly (line 148). I suggest that the PublicAPI call the actions and NOT the underscore versions to make it easier to override.

cibernox:
@cah-briangantzler About the first point, At this point, either return true or false are valid options, since this is new api, but I think that returning true like ember does for make actions bubble to routes is more an exception than a rule. In general in javascript you need to explicitly opt out for the default behaviour either using `event.preventDefault()` or returning `false` (like jquery does for event handlers). I tried to use `preventDefault` but that is not possible with `input` events.


I'm going to look at the second issue now. I have allocated some time to work on the component this morning (easier animations coming :))

miguelcobain:
I'm totally :+1: for this, especially if it is backwards compatible.

cibernox:
I just introduced a breaking change that forced me to consider the new release part of the 0.10 cycle, so there is no reason to hold this. A new beta period has begun. Let's see if this has any caveat.

PR #368, opened 2016-03-14
Fix concatenated class CPs

Dremora:
In `power-select.js`, `concatenatedTriggerClasses` and `concatenatedDropdownClasses` didn't depend on `triggerClass` and `dropdownClass` correspondingly, which basically made it impossible to have dynamic values for those properties.


cibernox:
Yep, makes total sense.

PR #383, opened 2016-03-19
Added acceptance tests that visit all public pages

ef4:
This kind of acceptance test is useful because it warns you if there are exceptions thrown while trying to render any of the public pages.


cibernox:
Great. Thx!

PR #384, opened 2016-03-19
Improve docs pages scroll handling

ef4:
This will cause every public page to start out scrolled to the top, and then when returning to an already visited route will return to the last remembered scroll position.

This makes moving forward and backward through the documentation pages feel nicer, because forward links scroll you to the top automatically, and backward links take you back to where you were when you clicked forward.


cibernox:
Thanks!

Issue #380, opened 2016-03-18
Click on delete add an empty box


GCorbel:
In Ember 2.4 and ember-power-select 0.10.0-beta.2. I have this route : 

```
import Ember from 'ember';

export default Ember.Route.extend({
  model() {
    var courses = this.store.findAll('course');
    return this.store.createRecord('newsletter', { courses });
  },
});

```

And this template : 

      {{#power-select-multiple options=courses selected=model.courses onchange=(action (mut model.courses)) as |course|}}
        {{course.name}}
      {{/power-select-multiple}}

It show all courses but, I click on the cross to delete an element, it add a black rectangle like this : 

![capture du 2016-03-18 08 30 59](https://cloud.githubusercontent.com/assets/1296201/13877859/cf9505c4-ece3-11e5-9da1-b37d36c2ca4f.png)

Is there a problem with promises?

GCorbel:
Solved by #386

PR #386, opened 2016-03-20
Convert to array when it's possible

GCorbel:
This fix issue #380 

With ember data, it's not possible to do `items.get('0')` to have the first value and it have to be converted to array. 

I can't do an integration test because [this line](https://github.com/cibernox/ember-power-select/blob/master/tests/integration/components/power-select/ember-data-test.js#L20) gives an error because `server` is undefined.

### DIFF
--- a/addon/components/power-select-multiple/trigger.js
+++ b/addon/components/power-select-multiple/trigger.js
@@ -21,8 +21,12 @@ export default Ember.Component.extend({
       if (e.target.dataset.selectedIndex) {
         e.stopPropagation();
         e.preventDefault();
+
+        let index = e.target.dataset.selectedIndex;
         let selected = this.get('selected');
-        this.get('select.actions.choose')(get(selected, e.target.dataset.selectedIndex));
+        let object = this.selectedObject(selected, index);
+
+        this.get('select.actions.choose')(object);
       }
     };
     if (isTouchDevice) {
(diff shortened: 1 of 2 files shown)


cibernox:
@GCorbel I've updated mirage and fixed the test suite so you can properly test this.

I think however that instead of `toArray` you should use objectAt
```js
if (!selected.objectAt) {
  selected = Ember.A(selected);
}
this.get('select.actions.choose')(selected.objectAt(e.target.dataset.selectedIndex));
```

cibernox:
@GCorbel Do you have a moment for rebasing and amend this PR?

Thanks

GCorbel:
Yep. I will do it right now!

GCorbel:
I rebased it but I'm not able to reproduce the issue on tests.

    $('.ember-power-select-multiple-remove-btn')[0].click();

Do not work.

I will check to night.

cibernox:
@GCorbel that is totally expected. This addon uses native events, not jquery events. Check how other tests do the same (`nativeMouseDown`)

GCorbel:
I did my work! Thanks for your help!

GCorbel:
Tests are failing. I'll fix it tomorrow.

GCorbel:
Now it works.

cibernox:
Thanks! I'll release a new beta very soon.

PR #390, opened 2016-03-21
Customized Options: fixed selection closes #389

### LINKED ISSUE #389: Custom Option-Elements broken
Since v0.10.0-beta.0 options with custom elements are broken.

Example:
```handlebars
{{#power-select options=someList selected=selectedItem onchange=(action 'someAction') as |item|}}
  <div class="special-class">
    <i class={{item.icon}}></i>
    {{item.text}}
  </div>
{{/power-select}}
```

If you click on the 'i'-Tag in the list, nothing happens.
After some investigation i found out, why it doesn't work:

[addon/components/power-select/options Line 19](https://github.com/cibernox/ember-power-select/blob/74202ee25bc189883edca130b75c76eb2eec9410/addon/components/power-select/options.js#L19)
```javascript
this.element.addEventListener('mouseup', e => {
  if (e.target.dataset.optionIndex) { // the clicked element 'i' doesn't have a data-option-index property
    this.get('select.actions.choose')(this._optionFromIndex(e.target.dataset.optionIndex), e);
  }
});
```

The clicked 'i' doesn't have a 'data-option-index'-Property, so the if gets passed by and nothing happens.

That could be solved with the little help of jquery:
```javascript
this.element.addEventListener('mouseup', e => {
  const optionIndex = Ember.$(e.target).closest('li').data('option-index').toString();
  if(!isEmpty(optionIndex)){
    this.get('select.actions.choose')(this._optionFromIndex(optionIndex), e);
  }
});
```

Maybe there is also another solution?

Haidy777:
see #389

### DIFF
--- a/addon/components/power-select/options.js
+++ b/addon/components/power-select/options.js
@@ -1,8 +1,6 @@
 import Ember from 'ember';
 import layout from '../../templates/components/power-select/options';
 
-const { get } = Ember;
-
 export default Ember.Component.extend({
   isTouchDevice: (!!self.window && 'ontouchstart' in self.window),
   layout: layout,
@@ -16,9 +14,9 @@ export default Ember.Component.extend({
     if (this.get('role') === 'group') { return; }
 
     this.element.addEventListener('mouseup', e => {
-      if (e.target.dataset.optionIndex) {
-        this.get('select.actions.choose')(this._optionFromIndex(e.target.dataset.optionIndex), e);
-      }
+      const optionItem = Ember.$(e.target).closest('[data-option-index]');
+      if (!optionItem) { return; }
+      this.get('select.actions.choose')(this._optionFromIndex(optionItem[0].dataset.optionIndex), e);
     });
     this.element.addEventListener('mouseover', e => {
       if (e.target.dataset.optionIndex) {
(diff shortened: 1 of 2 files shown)


addon/components/power-select/options.js
```diff
@@ -16,8 +16,9 @@ export default Ember.Component.extend({
     if (this.get('role') === 'group') { return; }
 
     this.element.addEventListener('mouseup', e => {
-      if (e.target.dataset.optionIndex) {
-        this.get('select.actions.choose')(this._optionFromIndex(e.target.dataset.optionIndex), e);
+      const optionIndex = Ember.$(e.target).closest('li').data('option-index').toString();
```
cibernox: Do you need the `toString`? 

What about...

```
let optionItem = e.target.closest('[data-option-index]');
if (!option) { return; }
this.get('select.actions.choose')(this._optionFromIndex(optionItem.dataset.optionIndex), e);
```

cibernox: I prefer this in case the custom html inside the options contains (God knows why) more nested `<li>`s elements.

cibernox:
Just one suggestion, to make sure this works even if there is nested `<li>`s in the markup.

cibernox:
Failures in travis seem related to npm doing npm things. Should be fine the next time.

Haidy777:
Well, didn't think about nested ```li```s. Good point ;)
```.toString()``` is only needed because ```.data``` returns a Number (in this case).

Just updated the commit.

cibernox:
Thanks!

cibernox:
Released a new beta

PR #396, opened 2016-03-22
Remove `opened` feature

cibernox:
[REMOVE FEATURE] The opened property (the only using double bindings instead of DDAU) has been removed. It was the cause of some errors due to race conditions in the bindings propagation. It is still possible to pass `initiallyOpened=true` to render a select already opened, but it is a one time property. It won't onpen/close the select when mutated nor will be mutated when the select is opened or closed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Master
 
+# 0.10.0-beta.4
+- [REMOVE FEATURE] The `opened` property (the only using double bindings instead of DDAU) has been
+  removed. It was the cause of some errors due to race conditions in the bindings propagation.
+  It is still possible to pass `initiallyOpened=true` to render a select already opened, but it is
+  a one time property. It won't onpen/close the select when mutated nor will be mutated when the
+  select is opened or closed.
 - [BUGFIX] Fix option highlighting when the use mouseovers in an element inside the `<li>`s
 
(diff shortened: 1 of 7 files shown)


cah-briangantzler:
I was using that in several places. Is there a proposed solution to opening the drop down programmatically? (Thanks for at least the initialRender)

Luckily since I subclassed and provided a openOnFocus, I can just focus the select and get it to open (reconsider a PR for openOnFocus?)

cibernox:
Sorry about this removal. I was tracking down a few a few bugs and I always ended in the same place, this property being set several times and calling `open` or `close`.

I'd say that in your use case, focus the component is a perfect replacement. Alternatively I could expose an action that is called when the component is added to the DOM (`didInsertElement`) that allows you to grab a reference to the public API, and from that point on use that object as a remote controller.

But this double binding thing was a source of many problems. I'd be happy to provide a replacement for your use case if the focus strategy doesn't work that doesn't involve double bindings. It was pretty nightmarish.

cah-briangantzler:
I like the idea of an action sending the public api. Would give access to both open and close.

However isOpen, highLighted and searchText would not be updated, so could lead to confusion. Would have to change the API to a constant object and update instead of creating new object.

cibernox:
That actually seems doable too. Keep the public API as a constant reference and update it.

PR #399, opened 2016-03-22
[TEST HELPER]: added clearSelected and removeSelectedMulOption test h…

snewcomer:
…elpers

Not sure how you feel about the try catch block, but it still throws an error and then additionally console warns if the css class path is incorrect.  I can take it out if you want!

### DIFF
--- a/test-support/helpers/ember-power-select.js
+++ b/test-support/helpers/ember-power-select.js
@@ -114,4 +114,24 @@ export default function() {
     }
 
   });
+
+  Ember.Test.registerAsyncHelper('removeMultipleOption', function(app, cssPath, value) {
+    const elem = find(`${cssPath} .ember-power-select-multiple-options > li:contains(${value}) > .ember-power-select-multiple-remove-btn`)[0];
+    try {
+      nativeMouseDown(elem);
+    } catch (e) {
+      console.warn('css path to remove btn not found');
+      throw e;
+    }
+  });
+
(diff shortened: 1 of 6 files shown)


test-support/helpers/ember-power-select.js
```diff
@@ -114,4 +114,24 @@ export default function() {
     }
 
   });
+
+  Ember.Test.registerAsyncHelper('removeSelectedMulOption', function(app, cssPath, value) {
```
cibernox: I'd probably call it `removeMultipleOption`. I find the "selected" redundant because it has to be selected to be removed.

tests/acceptance/helpers-test.js
```diff
@@ -117,4 +117,45 @@ test('selectSearch helper searches in the given multiple select closed', functio
...
\ No newline at end of file
+});
+
+test('removeSelectedMulOption removes selected option', function(assert) {
+  visit('/helpers-testing');
+
+  selectChoose('.select-choose-onopen-multiple', 'three');
+  selectChoose('.select-choose-onopen-multiple', 'four');
+  andThen(function() {
+    assert.equal(currentURL(), '/helpers-testing');
```
cibernox: I don't see any need of asserting the current url.

cibernox:
Added two minor comments. About the try/catch, seems fine.

Can you just add some docs in http://www.ember-power-select.com/docs/test-helpers ?

PR #400, opened 2016-03-23
Test Helpers: removeMultipleOption and clearSelected

snewcomer:
with tests and docs updated.  Had to reset one test b/c test run was failing on one commit after wait helper commit.


snewcomer:
Fail perhaps due to the wait helper that was added???

cibernox:
@snewcomer Yep, my bad. Fixed.

cibernox:
Thanks!

PR #401, opened 2016-03-24
[RFC] Allow to have placeholders with html

montoias:
This enables the possibility of doing stuff like:

```
placeholder='Current month is <span style="font-weight:bold”>Feb.</span>’
```

Which migrates the placeholder from being a **string** to **HTML**


cibernox:
Even if it is a bit more involved, I think that having html placeholder needs to be achieved passing a custom component. In the multiple-select the placeholder is passed to the `<input placeholder={{placeholder}}>`, so in that case it can't be html.

And since strings with html can't just be trusted, the passed string would need to be a SafeString, already making the api less nice to use. Not to bad tho.

Ideally being able to pass multiple blocks to a component would solve this issue, but is a feature that has been in the backlog for long in ember and I don't expect it to be finished soon.

Would passing an component be too cumbersome for you use case?

montoias:
I agree with what you mentioned, If we could pass multiple blocks it would definitely be much easier.

What worries me It's the fact that one would have to keep it in his code-base and maintaining it as future versions ember-power-select are released,  since the trigger redefinition component would either be:

- A simple placeholder (ignoring all other logic and **new that might appear**)
- Keep all properties and the changing only the placeholder (which will have a lot of dup logic).

My suggestion, if you would like, is having a placeholderComponent, that would default for the current span. This would allow to only customize what is really needed. This would allow for more granular component redefinition.

Blackening999:
@montoias @cibernox 
hey guys!

This issue become crucial on one of my projects, would be happy to see it merged?
Or at least some kind of advice to be given?
Maybe put a component into a placeholder property?

```hbs
placeholder=(component 'placeholder-component)
```

Thanks in advance,
Best

cibernox:
Actually I don't even remember why this didn't get merged.

@Blackening999 If you wish to have this features you could continue from this PR. I think that it's a very good idea. Seems like a useful feature.

montoias:
I can rebase this PR if you're interested in merging @cibernox

cibernox:
@montoias Please do. I don't recall having any objection to this feature. I like it.

cibernox:
Thanks. I'll review this deeper this afternoon, but looks good.

cibernox:
Thanks!

cibernox:
@Blackening999 @montoias Released in 1.4.0. Docs in http://www.ember-power-select.com/docs/the-trigger

Blackening999:
@cibernox awesome!
I hadn't got a time to check or continue due to tight schedule

@montoias you're awesome! :)

montoias:
@Blackening999 You're welcome 💯

Issue #402, opened 2016-03-24
Mouseover causing select options to jump to top when using block options. 


AndrewMoore10:
v0.10.0-beta.5
I have a block for my select options

`<div class="idcard">
  <div class="well">
    <div class="picture">
      <span class="verthelper"></span><img />
    </div>
    <div class="details">
      <p>Full Name</p>
      <p>email@example.com</p>
      <p>EID</p>
    </div>
  </div>
</div>`

When I mouse over the options, the box jumps to the top with the following error in console:

> options.js:32 Uncaught TypeError: Cannot read property 'dataset' of undefined(anonymous function)

AndrewMoore10:
Submitted PR to fix the console error. 

The jump to top issue was believed to be caused by my nested block being 'float: left;'

PR #403, opened 2016-03-24
added check for 0 index item on optionItem

moeabm:
added check for 0 index item on optionItem for mouseover and mouseup
https://github.com/cibernox/ember-power-select/issues/402

### DIFF
--- a/addon/components/power-select/options.js
+++ b/addon/components/power-select/options.js
@@ -15,12 +15,12 @@ export default Ember.Component.extend({
 
     this.element.addEventListener('mouseup', e => {
       const optionItem = Ember.$(e.target).closest('[data-option-index]');
-      if (!optionItem) { return; }
+      if (!optionItem || !(0 in optionItem)) { return; }
       this.get('select.actions.choose')(this._optionFromIndex(optionItem[0].dataset.optionIndex), e);
     });
     this.element.addEventListener('mouseover', e => {
       const optionItem = Ember.$(e.target).closest('[data-option-index]');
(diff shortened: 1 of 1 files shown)


addon/components/power-select/options.js
```diff
@@ -15,12 +15,12 @@ export default Ember.Component.extend({
...
     this.element.addEventListener('mouseup', e => {
       const optionItem = Ember.$(e.target).closest('[data-option-index]');
-      if (!optionItem) { return; }
+      if (!optionItem || !(0 in optionItem)) { return; }
       this.get('select.actions.choose')(this._optionFromIndex(optionItem[0].dataset.optionIndex), e);
     });
     this.element.addEventListener('mouseover', e => {
       const optionItem = Ember.$(e.target).closest('[data-option-index]');
-      if (!optionItem) { return; }
+      if (!optionItem || !(0 in optionItem)) { return; }
```
cibernox: I've never seen this construction for checking if a value is zero. Is there any difference with the more common `!optionItem && optionItem !== 0`? If there is none, I'd say that I prefer the `!==` one, for familiarity.

AndrewMoore10:
I suppose `!optionItem[0]` would be equivalent.

addon/components/power-select/options.js
AndrewMoore10: Checking for existence of the zero index element. I suppose !optionItem[0] would be equivalent.

cibernox:
Thanks!

PR #406, opened 2016-03-29
fix deprecation warning for Ember 2.5.0

gertjanwytynck:
Fixed deprecation warning for Ember 2.5.0.

See here -> https://github.com/emberjs/ember.js/pull/12303

### DIFF
--- a/addon/components/power-select.js
+++ b/addon/components/power-select.js
@@ -155,7 +155,7 @@ export default Ember.Component.extend({
         isOpen: dropdown.isOpen,
         highlighted: this.get('highlighted'),
         searchText: this.get('searchText'),
-        actions: Ember.merge(ownActions, dropdown.actions)
+        actions: Ember.assign(ownActions, dropdown.actions)
       };
     }
     return {};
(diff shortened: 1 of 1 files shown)


miguelcobain:
What do we do with previous 2.5 Ember versions?
We'll probably need to have a polyfill for this polyfill...

urbany:
Maybe this is enough:
```javascript
const assign = Ember.assign || Ember.merge;
```
What do you think @miguelcobain ?

miguelcobain:
Since `Ember.assign` is `undefined` in previous Ember versions, it should work.
I would put it as an `import` at the start of the file (after other `import`s).

urbany:
@miguelcobain I agree. @gertjanwytynck can you make this change to your PR?

gertjanwytynck:
@urbany and @miguelcobain, good idea. Fixed?

miguelcobain:
Everything's green now. I think this can be merged @cibernox ?

cibernox:
Just woke up. Looks perfect.

Issue #407, opened 2016-03-29
When hovering over a disabled list-item there is a console error.


gertjanwytynck:
`Error: options.js:32 Uncaught TypeError: Cannot read property 'dataset' of undefined`

This is also on the demo page: http://www.ember-power-select.com/docs/the-list (bottom)

Any tips for a solution?

Haidy777:
Looks like that error is produced by this line: https://github.com/cibernox/ember-power-select/blob/master/addon/components/power-select/options.js#L24 .

I think a slightly altered if should fix it.

cibernox:
Thanks for reporting, this seems fixed now.

PR #409, opened 2016-03-29
Fix disabled field

gertjanwytynck:
See issue #407.


cibernox:
I think that this has been fixed by #403 
Can you confirm?

cibernox:
I've checked it myself,and added a test harness to verify avoid future regressions.

Thanks for this anyway!

PR #414, opened 2016-03-30
updated troubleshooting to use dropdownClass

toobulkeh:
updated troubleshooting guide to reflect the latest dropdownClass option, as `class='in-modal'` no longer solves this problem.


cibernox:
Good catch! Thanks!

Issue #415, opened 2016-03-30
Selected option doesn't transfer to blue-highlight?


toobulkeh:
Great library! Just switched over to it from xselect as xselect has no support for promises.

It took me a while to figure things out, but so far so good!

The first visual problem I have is when the selected option doesn't match the highlighted blue in the dropdown.

In this example we have a Model "Account" that has a list of users and a single owner:
`account.js`
```javascript
export default DS.Model.extend({
  owner: DS.belongsTo('user', {async: true}),
  users: DS.hasMany('user', {async: true}),
  ....
```
then a component that is a form consuming ember-power-select
`account-form.hbs`
```hbs
<div class="form-group">
  <label>Account Owner</label>
  {{#power-select
    dropdownClass='in-modal-dropdown'
    options=account.users
    selected=account.owner
    searchEnabled=false
    onchange=(action 'chooseOwner')
    as |user|}}
    {{user.fullName}}
  {{/power-select}}
</div>
```

I'm not exactly sure why this happens, but this is how it renders on first click:
![screen shot 2016-03-30 at 4 35 36 pm](https://cloud.githubusercontent.com/assets/1342064/14156829/6356e3ca-f696-11e5-84e2-9f9cb2058e85.png)

Some ideas that I think it's related to are:

1. It's in a modal. I only change the z-index though...
2. Both the options and selected are promises and there's [this currently open issue](https://github.com/cibernox/ember-power-select/issues/404), aptly numbered 404 :smiley_cat: 
2. I'm still not very skilled at finding when promises are resolved, so maybe account is still a promise as well?

Now that I think about it, I am using a custom action, because the mut shortcut didn't work... because #404... so is this yet another side effect that would be solved by the selected promise being resolved?

cibernox:
Yeah, this issue is most likely related. I'll get #404 fixed a few days after emberconf (i'm travelling at the moment). I have to make the `selected` property promise-aware as the `options` property.

cibernox:
@toobulkeh I think that I've fixed this. Try the beta.7 and let me know if that is not the case.

I'm optimistically closing this

toobulkeh:
yup, that was it! Thanks for the fast work!

PR #420, opened 2016-04-01
Make selected promise aware

### LINKED ISSUE #404: Selected wait for Promise
Great plugin but I am having an issue that I cannot figure how to solve. I have a dropdown that is returning a promise so when the page loads, the selected option is not getting set because the promise has not been fulfilled yet

// Route
```
  model(params) {
    return this.store.findRecord('employee', params.employee_id);
  },

  setupController(controller, model) {
    this._super(controller, model);

    controller.set('setAction', 'saveEmployee');
    controller.set('providerList', this.store.findAll('provider'));
  },

  renderTemplate() {
    this.render('admin/employeeform');
  },

  actions: {
    saveEmployee(saveEmployee) {

      saveEmployee.save().then(() => this.transitionTo('admin.employees'));
    }
  }
```

// Template

```
    {{#power-select
      selected=model.provider.content
      options=providerList
      onchange=(action (mut model.provider))
      as |provider|}}
        {{provider.name}}
      {{/power-select}}
```

// Model

```
export default DS.Model.extend({
  name: DS.attr('string'),
  phone: DS.attr('string'),
  email: DS.attr('string'),
  notes: DS.attr('string'),
  provider: DS.belongsTo('provider')
});
```

When the page loads and I click the dropdown, I can see the options. What is not working is on page load, the provider is not being selected. Thoughts?

cibernox:
Closes #404 #415

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Master
 
+- [FEATURE] The `selected` option can now also be a promise. When it's a promise, the component
+  won't have a have any selection (the trigger will be empty, no option of the list will be
+  highlighted) until trat promise resolves. Once it resolves, the triger and the highlighted
+  option will update.
 - [BUGFIX] Fixed bug after event-delegation refactor where mouseovering the list itself
   (which happens when options are disabled) throwed an error.
 - [BUGFIX] Disabled select shouldn't be clearable even if `allowClear=true`.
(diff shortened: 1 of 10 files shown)


miguelcobain:
I think there's a bit of an edge case we need to handle.
The `selected` promise can change *before* the previous one resolves.
We need to test that the selected promise is still the same when a promise resolves. Otherwise, we may be silently overwriting values.
ember-cli-selectize and Ember.Select did have this check, iirc.

PR #422, opened 2016-04-01
make sure we do not overwrite new selected promise values

miguelcobain:
see https://github.com/cibernox/ember-power-select/pull/420#issuecomment-204337797


cibernox:
Yep. Makes sense.

cibernox:
Thanks. Good eye 👀

miguelcobain:
I am watching you

PR #423, opened 2016-04-02
pass horizontalPosition through basic-dropdown component

Pasikonik:
exactly that


cibernox:
Looks good. 
Can you just add a test to check that the dropdown has the proper class?

Pasikonik:
added

cibernox:
Thanks!

Issue #293, opened 2016-02-06
RFC - Upcoming a11y design


cibernox:
After some comments in other issues, I've decided to gather together all the a11y design and best practices. 

I've also found [this example](https://www.w3.org/TR/wai-aria/roles#combobox) to be particularly enlightening. 

NOTE: The image deliberately omit the roles of the options of the list because seems to be a consensus that are `role=option aria-selected=true/false aria-current=true/false aria-disabled=true/false`. This focuses on what wires the trigger, the searbox and the list of options together only.

Starting with the most basic component, Ember Basic Dropdown, the a11y is relatively simple.

<img width="706" alt="basic-dropdown-a11y" src="https://cloud.githubusercontent.com/assets/265339/12863279/b067a564-cc6b-11e5-96a1-7e1a80772533.png">

On top of this we build the power-select and power-select-multiple

<img width="689" alt="single-select-a11y" src="https://cloud.githubusercontent.com/assets/265339/12863320/c71ad934-cc6b-11e5-976d-2fa6902fc403.png">

<img width="627" alt="multiple-select-a11y" src="https://cloud.githubusercontent.com/assets/265339/12863323/d0988aec-cc6b-11e5-85d3-3e9d370378b1.png">

I'm also omitting `aria-description`, `aria-label`.

cc @jmacqueen @martndemus

In case this is the path taken:

- [x] The options list has `role=listbox`.
- [x] Groups have `role=group`
- [x] Each option has `role=option`.
- [x] The selected option has `aria-selected=true`.
- [x] The highlighted option has `aria-current=true`.
- [x] The disabled options has `aria-disabled=true`.
- [x] Searchbox of the single component has `role=combobox`.
- [ ] Searchbox of the single select has `aria-expanded=true`, `aria-owns=id-of-the-listbox` and `aria-autocomplete=list`.
- [ ] Searchbox of the multiple select has `aria-expanded=true`, `aria-owns=id-of-the-listbox` and `aria-autocomplete=list`. **aria-expanded** has to be false/true depending on if the options are visible?
- [ ] Trigger has `role=button`
- [x] Trigger `aria-owns=<id-of-dropdown>`
- [ ] Trigger has `aria-pressed=true/false`. **Should we remove `aria-expanded` then?**
- [ ] Trigger get `aria-disabled=true` when passed `disabled=true` to the component.
- [ ] Trigger has `aria-haspopup=true`.
- [ ] Accepts `aria-labelledby` (applies to the trigger).
[...]

PR #433, opened 2016-04-05
Pass WAI-ARIA states into dropdown

jmacqueen:
This will pass the following WAI-ARIA relationship and widget attributes through to `basic-dropdown` in an attempt to complete cibernox/ember-power-select#293

I have a thing about alphabetizing long property lists to help out when exploring or debugging, I hope that doesn't cause too much trouble. Other than alphabetizing, these are the only actual additions to the templates.

`power-select` attribute  | WAI-ARIA attribute in DOM |  Note
--------------------------------|--------------------------------------|----------
`ariaDescribedBy`   | `aria-describedby`
`ariaInvalid `            | `aria-invalid`
`ariaLabel `              | `aria-label`                     | This one still needs to be added to `basic-dropdown`, but inclusion here ahead of time doesn't break anything
`ariaLabelledBy `     | `aria-labelledby`
`required`                 |`aria-required`

I didn't include tests because I'm a bad person. And I didn't want to take the time on the off-chance these changes don't go in :smile:


package.json
```diff
@@ -67,7 +67,6 @@
     "ember-cli-babel": "^5.1.5",
     "ember-cli-htmlbars": "^1.0.1",
     "ember-basic-dropdown": "^0.9.5-beta.10",
-    "ember-hash-helper-polyfill": "^0.1.0",
```
cibernox: I agree that I not longer need this polyfill, but it concerns me that [this test](https://github.com/cibernox/ember-power-select/blob/master/tests/integration/components/power-select/customization-with-compoments-test.js#L113) is not failing in 1.13 in travis.

Nothing for your to change, just a note to myself.

cibernox:
I don't mind the alphabetically order. I started grouping properties "conceptually", but clearly there was a moment where that stopped scaling :sweat_smile: . This is better now.

This **does** need tests tho (and probably some work in ember-basic-dropdown itself).

cibernox:
Tests in ember-basic-dropdown and ember-power-select overlap a bit, but I try to have tests in EPS for behaviours that are anyway tested in EBD because sometimes they depend of EPS forwarding the proper properties, so in this case the duplication has a purpose.

jmacqueen:
I've ended up duplicating in similar ways (and for similar reasons) in the widget set I'm making at work. It's helped to catch refactoring problems when I thought I was safely changing something isolated in one component...and it ends up I was wrong!

Especially considering the case of integration tests, in my opinion, it's well worth having all the relevant tests for the EPS interface in the EPS repo...even if it means tests aren't technically as DRY as they could possibly be.

jmacqueen:
Tests added. I went ahead and put in tests for `aria-label` that will fail until `ariaLabel` is added to `cibernox/ember-basic-dropdown` and the updated version included.

cibernox:
@jmacqueen mmm... why travis is not running for the last commit? Also, should we update ember-basic-dropdown first?

cibernox:
Rebase over master to get the version of EPS with support for aria-label.

cibernox:
Thanks for all the a11y work. 

This starts to look complete.

Issue #440, opened 2016-04-07
Removing the ul in ember-power-select-multiple


karellm:
Hi,

I noticed that selected options are wrapped in a `ul.ember-power-select-multiple-options`. This causes the input where we type to be pushed on a new line when selected options don't fit on a single line:

![](https://s3.amazonaws.com/f.cl.ly/items/2W1f1L1c2L1i3T2r3G3T/Image%202016-04-07%20at%201.12.05%20PM.png?v=93f8d866)

As reference, here is how selectize does it:

![](https://s3.amazonaws.com/f.cl.ly/items/3d0z0U1d2G343p2f1R3n/Image%202016-04-07%20at%201.11.05%20PM.png?v=7947784d)

Is there a reason for that? Are you open to a PR?

karellm:
@cibernox This is blocking us in our application, can you tell me whether I should spend time on a PR or hack the solution locally?

cibernox:
I think I can move the input inside the `<ul>`. I don't know how bad it is from an accesibility point of view, but probably some `role=*` can fix it. Will be fixed today

cibernox:
This should be fixed now in the master branch.

karellm:
Thanks that did the trick!

PR #458, opened 2016-04-18
Avoid breakline in multiple selects

cibernox:
This has been done by making the input a sibling of the options, inside the `<ul>`.
I'm not sure if this is an a11y no-no.

It might be. Check.


cibernox:
Fixes #440

PR #462, opened 2016-04-21
Add `overflow: hidden` to trigger CSS

grapho:
`ember-power-select-trigger` (while in single mode).  Needs to have `overflow: hidden` in order for the text-overflow property to make it's ellipsis.


cibernox:
Isn't that already handled by this line: https://github.com/cibernox/ember-power-select/blob/master/app/styles/ember-power-select.scss#L21 ?

grapho:
gah! you are right... i have an out of date version.

cibernox:
Confirm that it works and if so then close 👍

grapho:
yeah it works! thanks! and sorry!

Issue #391, opened 2016-03-21
Add integration test helpers documentation to website


blimmer:
The acceptance test helpers are well documented, but the integration test helpers aren't exposed on the website.

snewcomer:
@cibernox I can take a wack at this if you want.  Basically two sections - Acceptance Test Helpers and Integration Test Helpers.  See a few typos on the helpers page as well I can fix.

cibernox:
@snewcomer Absolutely.

I never intended to make those integration helpers public API, just the acceptance ones (which BTW, are tested so I can be sure I don't break them). 

But I guess that once they are added, there is no harm in telling people that they exist. Just add a small disclaimer that they are unlikely to change, but there is a very small chance that they do.

PR #464, opened 2016-04-22
DOCS - Integration Tests

snewcomer:
Issue: https://github.com/cibernox/ember-power-select/issues/391

### DIFF
--- a/tests/dummy/app/templates/public-pages/docs/test-helpers.hbs
+++ b/tests/dummy/app/templates/public-pages/docs/test-helpers.hbs
@@ -98,6 +98,15 @@
 <br>
 <h2>Integration Tests</h2>
 
+<p>
+  Test helpers must be imported at the top of your integration test.  The basic functionality of these helpers are unlikely to change, but may experience minor revisions in the future.  
+</p>
+<p><code>import { typeInSearch, clickTrigger } from '../../../helpers/ember-power-select'</code></p>
+<p>
+  Also note that you could extract the code for these test helpers into your own helpers file was well.
+</p>
+<p><code>import { typeInSearch, clickTrigger } from 'my-app/tests/helpers/test-helpers'</code></p>
+
 <h3><code>clickTrigger(scope, options = {})</code></h3>
 
 <p>
--- a/tests/dummy/app/templates/public-pages/docs/test-helpers.hbs
+++ b/tests/dummy/app/templates/public-pages/docs/test-helpers.hbs
@@ -1,8 +1,11 @@
 <h1 class="doc-page-title">Test helpers</h1>
 
+<br>
+<h2>Acceptance Tests</h2>
+
 <p>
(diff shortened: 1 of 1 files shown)


cibernox:
Thanks! I'll deploy in a moment

et:
Any chance we could get more available for integration tests? In particular `selectChoose` from the acceptance tests would help a lot.

et:
For context, this would help me when using something like `onchange=(action "someAction")` where `someAction` is doing extra stuff. I need a way to trigger that action to occur.

snewcomer:
@et 
There is a helper:
```
import { typeInSearch, clickTrigger, nativeMouseUp } from '../../../helpers/ember-power-select';
clickTrigger();
nativeMouseUp('.ember-power-select-option:eq(3)');
```
I do something like this as well w/ jquery.

`$(".ember-power-select-option:contains('selectMe')").mouseup()`

snewcomer:
Let me know if this isn't exactly what you needed.  Based on your comment, it seems like this `nativeMouseUp` could go in the docs as well.

et:
Thanks @snewcomer - are you copying/pasting the contents of that helper into your project or is there a way to make this perform a lookup?

snewcomer:
@et either into your own helper file and copy and paste the functions from here.
https://github.com/cibernox/ember-power-select/blob/master/test-support/helpers/ember-power-select.js#L30
or import it
import { typeInSearch, clickTrigger, nativeMouseUp } from '../../../helpers/ember-power-select';

et:
Great! I think it would make sense to explicitly say that on [this page](http://www.ember-power-select.com/docs/test-helpers). Thoughts?

snewcomer:
About the import it does.  Just not about the nativemouseup function.

et:
Sorry, I was unclear. I'm referring to the fact that you need to copy/paste the functions. As someone who is new to the project, it was not obvious where these functions are derived from and that you actually needed to copy them over.

snewcomer:
Sry a little confused, but that might be because I just woke up :) .  Under the integration tests section, it says these helpers must be imported.  In acceptance tests, the whole app is booted up, so you get these functions for free.  Copying and pasting them over if you so decide ensures that any api changes to the helpers in the future doesn't all of a sudden break your build.  Just imagine if you had a project with thousands of tests that only built when your tests pass and there was an api change to the helper.  It may break your build if you aren't careful.  Most people can just import the helpers from power select as long as the project follows the upgrade path as Ember upgrades.

et:
It might be useful then for the install process to copy these test helpers into your project. As I understand it, that's pretty common for addons.

cibernox:
@et I'm also confused. Why would you want to copy the helpers into your app if you can import them?

et:
TIL - ember addons can implicitly drop themselves into your tests helper directory. Sorry about this, I wasn't aware this was a thing. This all makes sense.

PR #467, opened 2016-04-26
Support passing in custom `groupNameComponent`

JackCA:
This is a pretty simple spike that allows you to pass in a  `groupNameComponent` option to a top-level `{{power-select}}` component.

 This component will be used instead of just rendering the plain text `groupName` in a `<span>`.

This would, for example, allow you to create icons for each of your groups or anything else that is a little more involved than the plain text.

If you give the go ahead, @cibernox, I will test and document this and finish out the pull-request.

Thoughts?

### DIFF
--- a/addon/templates/components/power-select.hbs
+++ b/addon/templates/components/power-select.hbs
@@ -59,6 +59,7 @@
       searchText=(readonly searchText)
       select=(readonly publicAPI)
       selected=(readonly resolvedSelected)
+      groupNameComponent=(readonly groupNameComponent)
       as |option term|}}
       {{yield option term}}
     {{/component}}
--- a/addon/templates/components/power-select/options.hbs
+++ b/addon/templates/components/power-select/options.hbs
@@ -7,7 +7,10 @@
   {{#if opt.groupName}}
     <li class="ember-power-select-group" aria-disabled={{opt.disabled}} role="option">
       {{#if groupNameComponent}}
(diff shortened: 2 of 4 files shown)


cibernox:
This week has been pretty busy, sorry for the delay.

My main "concern" is that rendering component in a list can slow down the rendering time if there is many. On the other hand, typically there is far less groups than options, so probably is not that bad.

I feel positive. My only suggestion is that the component should probably reiceive more data. At least this:

```hbs
{{component groupNameComponent 
  option=opt 
  index=index
  select=(readonly select)
  selected=(readonly selected)
  highlighted=(readonly highlighted)}}
```

By example one use case that comes to mind is render the group name in a different color if the selected element is one of its options.

JackCA:
@cibernox awesome! I think you're use-case is a great example as well.

Yes, there could be some overhead in rendering more components but I think you're right that it should be relatively small.

I will implement adding that suggested data and will ping when it's done!

cibernox:
@JackCA life is going back to normal after holidays. There was some changes that makes this unmergeable, but the general idea is good. 

Go ahead and finish it if you want.

JackCA:
@cibernox great to hear! Yes, I'm interested in writing tests and proper docs but maybe a stranger's +1 would help motivate that?

I'd like to make sure this isn't just something applicable in my use case so if anyone is interested in this please +1 the original comment :)

les2:
+1

cibernox:
Closing for now, let me know if you want to reopen this.

PR #472, opened 2016-04-27
Remove jscs warnings

yethon:
Removes jscs warnings for anyone using a jscs linter such as `ember-suave`

### DIFF
--- a/test-support/helpers/ember-power-select.js
+++ b/test-support/helpers/ember-power-select.js
@@ -5,14 +5,14 @@ import Ember from 'ember';
 function typeText(selector, text) {
   let $selector = $(selector);
   $selector.val(text);
-  let event = document.createEvent("Events");
+  let event = document.createEvent('Events');
   event.initEvent('input', true, true);
   $selector[0].dispatchEvent(event);
 }
 
 function fireNativeMouseEvent(eventType, selectorOrDomElement, options = {}) {
   let event = new window.Event(eventType, { bubbles: true, cancelable: true, view: window });
-  Object.keys(options).forEach(key => event[key] = options[key]);
+  Object.keys(options).forEach((key) => event[key] = options[key]);
   let target;
   if (typeof selectorOrDomElement === 'string') {
     target = Ember.$(selectorOrDomElement)[0];
@@ -31,7 +31,7 @@ export function nativeMouseUp(selectorOrDomElement, options) {
 }
 
 export function triggerKeydown(domElement, k) {
-  var oEvent = document.createEvent("Events");
+  let oEvent = document.createEvent('Events');
   oEvent.initEvent('keydown', true, true);
   $.extend(oEvent, {
     view: window,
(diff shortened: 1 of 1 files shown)


cibernox:
Can you add ember-suave to the project itself, so no new violations are introduced?

yethon:
Absolutely!
(Also, looking into travis build failure.  Tests pass locally...)

cibernox:
If it fail only in some versions of ember but passes in others it might be a random failure I haven't had time to track. Start by retrying the failed build.

If it fails in all versions of ember-try, then is probably a legitimate error

yethon:
It looks like there's a known solution for the build failure: [ReferenceError: Can't find variable: Symbol](https://til.hashrocket.com/posts/b4c0e5d1a4-cant-find-variable-symbol), however that solution introduces a new build problem. Still investigating...

cibernox:
That error is weird. I merged two pull requests today and CI was green. I'm not aware of anywhere where I use symbols, and this addons has very few dependencies. 

I'll investigate it too.

test-support/helpers/ember-power-select.js
```diff
@@ -67,7 +67,7 @@ export default function() {
   const isEmberOne = Ember.VERSION.match(/1\.13/);
 
   Ember.Test.registerAsyncHelper('selectChoose', function(app, cssPath, value) {
-    const id = find(cssPath).find('.ember-power-select-trigger').attr('id').match(/ember-power-select-trigger-ember(\d+)/)[1];
+    const [,id] = find(cssPath).find('.ember-power-select-trigger').attr('id').match(/ember-power-select-trigger-ember(\d+)/);
```
cibernox: Maybe this kind of destructuring is the one that compiles to someting that uses `Symbol.iterator`?

yethon: I suspect that's the case....

cibernox: Phantomjs is a PITA

cesarizu: You can try with `ember-polyfill-for-tests`. It might solve those problems.

yethon: Thanks @cesarizu, I'll try that.

yethon:
@cibernox 
It looks like if we want to do array destructuring, (which I think we all do), we need a polyfill for Symbol. Looks like this polyfill gives us what we need.

However, considering the bigger picture - if we do this, it means all your users will need that polyfill. Is that reasonable? We already need to update our phantom version if we use power-select, so additional tweaks may not be a big deal.

I started to look at applying jscs checks (via ember-suave) to your entire project. Does that still interest you?  There are quite a few changes, but they are straightforward and I could submit changes incrementally before applying ember-suave to everything.

cibernox:
@yethon Yes, It would be nice to have some tooling enforcing a styleguide. 

However, about the destructuring, I found this (https://babeljs.io/docs/usage/caveats/). Apparently only array destructuring needs `Symbol.iterator`. I think that would be simpler to just avoid array destructuring, because I think that it would force people supporting IE 9 and 10 to include the polyfill (which I still think they should, but I don't want to be the cause)

yethon:
Fair enough.  I think I know a few alternative (sneaky) ways to tweak these lines that will silence jscs and won't require array destructuring... stay tuned.

test-support/helpers/ember-power-select.js
miguelcobain: To workaround jquery destructuring usually convert

```js
let [elem] = this.$('.btn');
// this errors at runtime but ember-suave doesn't complain
```

to

```js
let elem = this.$('.btn').get(0);
// this doesn't error at runtime neither ember-suave complains
```

yethon: Thanks @miguelcobain .

yethon:
@cibernox All warnings are gone, all tests pass....

cibernox:
@yethon Thanks!

Issue #408, opened 2016-03-29
Custom Option-Elements touch-events broken


Haidy777:
Since v0.10.0-beta.0 options with custom elements seem to be broken on touchy devices.

Might be a similar problem we had in #389? I couldn't find the problem yet, mabye someone else already did?

cibernox:
Labelled as bug, I'll check it after emberconf. Can you provide a failing test to help fixing it and prevent future regressions?

Haidy777:
Sure, I will try to add one tomorrow.

Haidy777:
Sorry I didn't reply any earlier. I've been pretty busy the last days.
Unfortunately I can't make the test running. Seems like they are not that easy to run on a windows environment?

cibernox:
@Haidy777 I never tried. The node ecosystem is not very windows-friendly in general, so this doesn't surprise me at all. 
What is failing? Phantomjs maybe?

Most people using node (and therefore using ember-cli) tend to use a linux VM (even if it's without UI). That is the only idea I can give you given that I never used ember-cli with windows.

Haidy777:
Sorry that it's been awhile since my last comment. I could track down the lines which are causing this problem.
As you can see in [addon/components/power-select/options.js](https://github.com/cibernox/ember-power-select/blob/master/addon/components/power-select/options.js#L43) it is the same kind of problem as in #389.

The touched element is not the one with the `options-index`. I've already got a solution, currently working on some tests. PR to come.

PR #474, opened 2016-04-28
Customized Options: fixed selection on touch devices

Haidy777:
See #408

### DIFF
--- a/test-support/helpers/ember-power-select.js
+++ b/test-support/helpers/ember-power-select.js
@@ -61,6 +61,27 @@ export function clickTrigger(scope, options = {}) {
   nativeMouseDown(selector, options);
 }
 
+export function nativeTouch (selectorOrDomElement) {
+  let event = new window.Event('touchstart', { bubbles: true, cancelable: true, view: window });
+  let target;
+
+  if (typeof selectorOrDomElement === 'string') {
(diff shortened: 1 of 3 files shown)


test-support/helpers/ember-power-select.js
```diff
@@ -61,6 +61,23 @@ export function clickTrigger(scope, options = {}) {
...
+  if (typeof selectorOrDomElement === 'string') {
+    target = Ember.$(selectorOrDomElement)[0];
+  } else {
+    target = selectorOrDomElement;
+  }
+  Ember.run(() => target.dispatchEvent(event));
+}
+
+export function touchTrigger () {
+  let selector = '.ember-power-select-trigger';
```
cibernox: One question. A thought "tap" is compound of a `touchstart` followed by `touchend` in which there wasn't any `touchmove` in between (otherwise is a drag).
I never heard of this `touchdown`, how does it work if I'm adding events for `touch{start|end}`?

Haidy777: @cibernox Well `touchdown` doesn't really exist, I don't know what drived me to that ;)
Just corrected it.

cibernox:
Thanks!

PR #478, opened 2016-05-02
break word on options

et:
@cibernox - what your thoughts are on this? This becomes an issue when you use power-select in a place that has a small container.

Before:
screen shot 2016-05-02 at 11 47 27 am

After:
screen shot 2016-05-02 at 11 47 35 am


ultimatemonty:
@et @cibernox I think you close this fella. It's implemented in the latest releases

cibernox:
Okay then 🎊

et:
Thanks for checking ❤️

PR #480, opened 2016-05-04
Add special class to loading message option

### LINKED ISSUE #479: No CSS class for loading message
All the specific option types such as "no matches message" and "search message" have a custom class except for the loading message. This has ended up with us using this less-than-ideal selector which happens to select only the loading message because we are using option types:

```
.ember-power-select-dropdown > .ember-power-select-options > .ember-power-select-option:first-of-type,
.ember-power-select-option--no-matches-message { }
```

kevinansfield:
closes #479
- makes the loading message option consistent with other special option elements by adding the custom class `.ember-power-select-option--loading-message`

### DIFF
--- a/addon/templates/components/power-select/options.hbs
+++ b/addon/templates/components/power-select/options.hbs
@@ -1,6 +1,6 @@
 {{#if loading}}
   {{#if loadingMessage}}
-    <li class="ember-power-select-option" role="option">{{loadingMessage}}</li>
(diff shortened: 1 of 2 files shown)


cibernox:
Makes sense to me. I'll merge once travis is green 👍

cibernox:
If you can, use master for now, I plan to dedicate some time to close issues this week and I'll cut a new version this weekend.

Issue #473, opened 2016-04-27
searchField and matcher incompatibility


karellm:
Hey,

I have a `power-select-multiple` of people of the form `{name, email}`. If I don't set the `searchField` and hit backspace, I get `Assertion Failed: `{{power-select-multiple}}` requires a 'searchField' when the options are not strings to remove options using backspace`. But when I set `searchField="email"`, then the matcher mis-behaves:

```js
recipientsSearchMatcher(user, term) {
  // I expect user to be the object {name, email}, but it is only the email
}
```

I run `0.10.1`.

Is that an expected behaviour or am I missing something? I would personally that the backspace doesn't throw an exception, it is fine to remove the item altogether without entering edit mode. If that's not an option, I would expect the matcher to still get the object and not just the email.

cibernox:
No, seems like a bug. I will be on holidays until the 4th, but I'll check it later.

cibernox:
If you can find a fix yourself, I should be able to merge from the phone :D

PR #482, opened 2016-05-04
Pass the value when using searchField with matcher

karellm:
@cibernox This should fix the problem in #473. I will obviously write tests and update the documentation before we can merge but since this is a breaking change for the API, I wanted to discuss it first.


cibernox:
I just came back from holidays. I'll check this later.

cibernox:
Superseded by #484

PR #484, opened 2016-05-05
[BUGFIX] Custom matchers receive the entire option even when used along with searchField

### LINKED ISSUE #473: searchField and matcher incompatibility

cibernox:
Closes #473

### DIFF
--- a/tests/integration/components/power-select/custom-search-test.js
+++ b/tests/integration/components/power-select/custom-search-test.js
@@ -3,7 +3,7 @@ import { moduleForComponent, test } from 'ember-qunit';
 import hbs from 'htmlbars-inline-precompile';
 import wait from 'ember-test-helpers/wait';
 import { typeInSearch, clickTrigger } from '../../../helpers/ember-power-select';
-import { numbers } from '../constants';
+import { numbers, countries } from '../constants';
 
 const { RSVP } = Ember;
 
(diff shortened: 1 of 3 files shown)


cibernox:
@karellm I think that your expectation in the bug report was actually accurate and a better idea that the idea you ended implementing in the PR.

My understanding, as yours in that issue, is that if the user provides a custom matcher to the component, that function should have access to the entire object and the search term. The `searchField` is only convenience to customize the default matcher and has no role here.

I actually consider the decission using the `searchField` for both searching and inferring the text to be displayed in the input when pressing backspace is probably not the best I've taken. Having that property fulfilling two different roles lead to this mess.

Check if this fixes your issue.

tests/integration/components/power-select/custom-search-test.js
```diff
@@ -551,3 +551,18 @@ test('BUGFIX: When the given options are a promise and a search function is prov
   typeInSearch("");
   assert.equal($('.ember-power-select-option').length, 20, 'There is 20 options again§');
 });
+
+test('BUGFIX: If the user provides a custom matcher, the that matcher receives the entire option even if the user also provided a searchField', function(assert) {
```
karellm: There is a typo `the that`

cibernox: 👀

karellm:
It looks great, should definitely fix my issue. I will implement when merged.

PR #489, opened 2016-05-09
Make it possible to disable groups

chrisgame:
Although groups cannot be clicked it would be great if they could be marked as disabled using the aria-disabled tag. Then styling could be applied depending on if the aria-disabled tag is true or false.

### DIFF
--- a/addon/utils/group-utils.js
+++ b/addon/utils/group-utils.js
@@ -80,7 +80,11 @@ export function filterOptions(options, text, matcher, skipDisabled = false) {
     if (isGroup(entry)) {
       let suboptions = filterOptions(get(entry, 'options'), text, matcher, skipDisabled);
       if (get(suboptions, 'length') > 0) {
-        opts.push({ groupName: entry.groupName, options: suboptions });
+        let groupCopy = { groupName: entry.groupName, options: suboptions };
+        if (entry.hasOwnProperty('disabled')) {
+          groupCopy.disabled = entry.disabled;
+        }
+        opts.push(groupCopy);
       }
     } else if (!skipDisabled || !entry.disabled) {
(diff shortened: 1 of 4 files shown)


cibernox:
@chrisgame Can you just add a small test that verifies this behaviour?

chrisgame:
ok, will do

chrisgame:
test added

addon/utils/group-utils.js
```diff
@@ -80,7 +80,7 @@ export function filterOptions(options, text, matcher, skipDisabled = false) {
     if (isGroup(entry)) {
       let suboptions = filterOptions(get(entry, 'options'), text, matcher, skipDisabled);
       if (get(suboptions, 'length') > 0) {
-        opts.push({ groupName: entry.groupName, options: suboptions });
```
cibernox: I don't get this line. Why do you need to modify this function?

chrisgame: Without it the disabled property that is present on the options that are passed in is lost so opt.disabled for a group in the template is undefined.

cibernox: I get it now. During filtering, a new objects is created.

I'd only suggest one change `{ groupName: entry.groupName, disabled: entry.disabled, options: suboptions }`. The generated group should have the same value as the original, including undefined. There is no need of coercing it to true.

cibernox: I'm not sure if this should be smart enough to not add the property `disabled` if the original doesn't have it.
```
let groupCopy = { groupName: entry.groupName, options: suboptions };
if (entry.hasOwnProperty('disabled')) {
  groupCopy.disabled = entry.disabled;
}
```

will make the test pass without changes. Do that.

PR #499, opened 2016-05-11
Use content editable to create autogrow input

cibernox:
Using a `<span>` with content-editable the autogrouth is automatic.

Probably requires some fixes in tests


cibernox:
This solution has too many edge cases. Closing in favour of a simpler approach

PR #502, opened 2016-05-12
Autogrow input using canvas-based text measurement.

### LINKED ISSUE #483: em-based width calculation is not reliable
It looks like the width of the `.ember-power-select-trigger-multiple-input` input is calculated from the length of its value. This is not reliable as you can see here:

In the recording, narrow letter work ok, while wide letters cause the input to be too small. Is there a reason why every letter adds `.5em` to the width? With webfonts, there is no correlation between a letter width and ems that I know of.

cibernox:
Closes #483


miguelcobain:
Genius!

karellm:
🎉  Thanks for that!

PR #504, opened 2016-05-13
Run before ember-cli-sri

### LINKED ISSUE #503: Should run before SRI Integrity Generation
Similar to this issue: https://github.com/miguelcobain/ember-paper/issues/224

If this addon generates its assets after ember-cli-sri has run, the integrity hash is wrong and causes this error:

Failed to find a valid digest in the 'integrity' attribute for resource

zacstewart:
[Fixes #503]

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,6 +72,7 @@
     "ember-truth-helpers": "^1.2.0"
   },
   "ember-addon": {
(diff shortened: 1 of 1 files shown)


cibernox:
I learned something new today 👍

cibernox:
Thanks!

PR #508, opened 2016-05-17
Make removeMultipleOption and clearSelected async friendly

### LINKED ISSUE #507: `removeMultipleOption` test helper does not use `wait` helper
Not sure if this is intentional or not, (if it is, I am sorry for the issue) but the `removeMultipleOption` test helper does not trigger the `wait` helper.

The following example is not working, without triggering the `wait` helper manually.
Changing the tags is making an async request to reload a list of items.

```javascript
removeMultipleOption('[data-autoid="search-select"]', 'master');

// Its not working without this line
wait(); 

andThen(() => {
  expect('.ember-power-select-multiple-option').to.have.size(0);

  // Fails here
  expect('[data-autoid="item"]').to.have.size(3); 
});
```

Let me know if this is intentional and I can open a pull request.
Thank you

mupkoo:
Makes the test helper function async friendly by adding wait calls.

Closes #507

### DIFF
--- a/test-support/helpers/ember-power-select.js
+++ b/test-support/helpers/ember-power-select.js
@@ -151,6 +151,7 @@ export default function() {
     const elem = find(`${cssPath} .ember-power-select-multiple-options > li:contains(${value}) > .ember-power-select-multiple-remove-btn`).get(0);
     try {
       nativeMouseDown(elem);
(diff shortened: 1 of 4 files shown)


cibernox:
Thanks!

PR #518, opened 2016-05-31
DOCS - Add "searchMessage" parameter description to api-reference table

kniteli:
I had to dig into the code to find this parameter. Figured I'd add it to the reference.


cibernox:
I don't know why the build failed but there is obviously no chance that this commit is the cause, so I'm merging it.

PR #523, opened 2016-06-01
Add null check for optionElement in scrollTo

ilucin:
When deleting an entry in multiple select field (with backspace) - scrollTo method can break when accessing offsetTop on optionElement which is null -> [bug reproducing video](https://monosnap.com/file/zjSVlECBVcwMi1BdCus0t7bpRXviXd)

This PR fixes my issue.


cibernox:
Can you check if if with 1.0.0-alpha.0 you can still reproduce this bug? The entire componen went through a total refactor, and after any refactor, old bugs might have gone and new bugs might have appeared.

ilucin:
Of course, I'll try it out with v1 alpha and let you know

ilucin:
Yep, this null issue is still here - testing with 1.0.0-alpha.10

cibernox:
I believe that this bug has been fixed now

ilucin:
@cibernox nop, it's still here, in the newest version

cibernox:
I have time to track it now. Working on it...

cibernox:
I tried to reproduce it but I couldn't. Any concrete step to get the error?

ayuk23:
It's difficult for me to recreate, but I've definitely had this issue come up numerous times.

cibernox:
Closing this very old PR.

PR #556, opened 2016-06-21
Fix keydown not being called in multiple-selects for A-Z

### LINKED ISSUE #555: Alpha2: onkeydown doesn't seem to trigger on [a-z] keypresses
Tried out the latest alpha (`v1.0.0-alpha.2`) and noticed that `onkeydown` does not trigger when pressing any of the letter keys.  It does trigger when pressing `enter`, directional arrow keys and anything on the number pad.

Is this is an expected change of behavior?

cibernox:
Closes #555


cibernox:
Closes #555

PR #568, opened 2016-06-27
Simplify keydown management

cibernox:


cibernox:
There is something broken in beta, but I don't see how this change can be the culprit. I think that it's a regression in ember.

PR #591, opened 2016-07-20
fix input box resizing for Internet Explorer

akatov:
need to compute the font property from base font properties on Internet Explorer


cibernox:
Looks like a more comprehensive implementation 👍

cibernox:
Thanks

PR #594, opened 2016-07-21
Allow disabling styles when not using Sass

rlivsey:
```
var app = new EmberApp(defaults, {
  'ember-power-select': {
    theme: false
  }
});
```


cibernox:
thanks

rlivsey:
No worries, cheers!

Issue #593, opened 2016-07-21
Prevent Keyboard opening on Android


Haidy777:
Is there a way to prevent the opening / focusing of the keyboard when the dropdown is opened (and has a Searchfield)?

Problem: If you have a list with a searchfield and only want to tap on entry three (which might be behind the keyboard) you can't do so unless you close the keyboard or scroll through the list.

cibernox:
Actually yes. That functionality is in the `beforeOptionsComponent` (https://github.com/cibernox/ember-power-select/blob/master/addon/components/power-select/before-options.js#L12). 

You can make a custom `beforeOptionsComponent` that extends from the default one 
```hbs
import BeforeOptionsComponent from 'ember-power-select/components/components/power-select/before-options';

export default BeforeOptionsComponent.extend({
  focusInput() {
    if (isMobile){
      return;
    }
    this._super(...arguments);
  }
});
```

That action is not entirely public API, but maybe I should provide a blessed path do do that.

cibernox:
Any idea how this could be configurable by the user without an very contrived option like `focusSearchOnOpen=false`?

Haidy777:
Well, I can't find something better ;) . ```focusSearchOnOpen``` Would, at least, describe what is does.

cibernox:
I'm thinking that with contextual component, it could be a configuration on the `beforeOptions` component.

```hbs
{{#power-select 
  options=options 
  selected=selected 
  beforeOptionsComponent=(component "power-select/before-options" autofocus=isDesktop) 
  as |opt|}}
  {{opt}}
{{/power-select}}
```

That way you don't need to create your custom component, just use the default customized with one property. It still adds one config option, but at least it lives where it's useful and the parent component doesn't need to worry about forwarding it. Beside, that wouldn't work in multiple select.

What do you think if this approach?

Haidy777:
That looks good to me, since its relatively easy to use and doesn't blow up the API with things that might be the lesser use case.

Also it shouldn't be that hard to implement, or am I missing a thing here?
(Should be a simple if here: https://github.com/cibernox/ember-power-select/blob/master/addon/components/power-select/before-options.js#L12)

cibernox:
Correct, a simple `if (this.get('autofocus')){  ...  }` there. 
Do you want to PR it? I han have it published today.

cibernox:
With a test, to make sure a refactor doesn't remove this functionality.
And one [ENHANCEMENT] entry in the changelog.

[...]

PR #596, opened 2016-07-22
Added optionally "autofocus" property to before-options component.

Haidy777:
See #593 for details.


cibernox:
Thanks! I'll release a new beta in a moment

PR #599, opened 2016-07-22
Allow customizing the search message component

rlivsey:
Adds `noMatchesMessageComponent` and `searchMessageComponent` options to customize the no-matches and search message components respectively.


cibernox:
As discussed in person, I'm in on the `searchMessageComponent` but not so much in the `noMatchesMessageComponent`, because the component accepts an `else` block where you can already render any component you want.

rlivsey:
👍 

I've removed the `noMatchesMessageComponent` and rebased so it just adds the `searchMessageComponent` now.

cibernox:
@rlivsey Tests are failing

cibernox:
Ok, disregard the failure. The components seems to be entirely broken in beta, no idea why. Merging.

PR #613, opened 2016-08-02
Use liquid-fire from newest commit as of today. Fixes failing tests in beta

cibernox:


cibernox:
Apparently it's not the only failure but either way, is an improvement.

PR #624, opened 2016-08-08
Check if app.options is available

mydea:
In an engine, the addon errors because `app.options` is `undefined`. This PR guards against this by checking if `app.options` is an object.


cibernox:
I'm back from holidays. This looks good. Merging.

PR #627, opened 2016-08-09
Add keyup

dustinspecker:
Adds support for keyup handler for the search input if you're interested.

We came across a need for it (or we could be overlooking something). Currently, we've setup a higher level event listener, but it would be great if a public API existed, so this PR makes that API.

Thank you for ember-power-select; it has really helped us a lot!

### DIFF
--- a/addon/components/power-select.js
+++ b/addon/components/power-select.js
@@ -308,6 +308,14 @@ export default Component.extend({
       return this._routeKeydown(e);
     },
 
+    // keyups handled by inputs inside the component
+    onKeyup(e) {
+      let onkeyup = this.get('onkeyup');
+      if (onkeyup) {
+        onkeyup(this.get('publicAPI'), e);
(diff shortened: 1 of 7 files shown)


dustinspecker:
So funny story. Turns out we need the `input` event handler instead of `keyup` on the search input.

If you'd like this pull request, definitely feel free to merge.

I am curious about your thoughts on exposing the `input` event. If this PR got merged, seems like a precedence would be set for always making a new PR to get a desired event handler. Are you set on only supporting `keydown` or is there a generic way we could expose all events to the API? We would be open to making a PR for the desired changes, bu wet want to make sure we'd be going down a path you would agree with.

Thank you for your time! We really appreciate your work.

cibernox:
@dustinspecker There is already an `oninput` action actually. It has been there for ages, but I forgot to document it.

Done.

PR #640, opened 2016-08-19
Check for isDestroyed when closing after choose action

osxi:
Fixes `vendor.js:29225 Uncaught Error: Assertion Failed: calling set on destroyed object: <frontend@component:power-select-search::ember1935>.placeholder = [object String]` error when closing after choosing an option with `power-select`.

Can provide a repro repo if you'd like but basically, it started with a controller that had a property like `showSomething` along with the following template:

```
{{#if showSomething}}
  {{#power-select}}...
{{else}}
   {{#power-select}}...
{{/if}}
```

While inside the first half of the `if` statement and choosing an option, the `else` is rendered and the error above is thrown.

This PR fixes that by checking for `isDestroyed` before actually closing the component via `publicAPI.actions.close(e);`.

### DIFF
--- a/addon/components/power-select.js
+++ b/addon/components/power-select.js
@@ -278,7 +278,7 @@ export default Component.extend({
       }
       let publicAPI = this.get('publicAPI');
       publicAPI.actions.select(this.get('buildSelection')(selected, publicAPI), e);
-      if (this.get('closeOnSelect')) {
+      if (!this.get('isDestroyed') && this.get('closeOnSelect')) {
         publicAPI.actions.close(e);
         return false;
       }


cibernox:
@osxi I'm curious, how come can the `choose` action be called on a destroyed select?

osxi:
@cibernox I'm coming up with a better description. I'm still investigating but there's a possibility that we're mutating something instead of using DDAU.

cibernox:
Even without DDAU, if the component is destroyed you can't open & click on it, or press enter/space. I don't see how this situation can happen under normal circumstances and I wonder what kind of flow can lead to this.

osxi:
@cibernox https://github.com/osxi/power-select-bug

The `placeholder` error can be avoided by commenting out https://github.com/osxi/power-select-bug/blob/master/app/components/power-select-search/component.js#L25-L29 and https://github.com/osxi/power-select-bug/blob/master/app/components/power-select-search/component.js#L39, which creates this error: `ember.debug.js:19699 Uncaught Error: Assertion Failed: calling set on destroyed object: <power-select-bug@component:power-select::ember389>.publicAPI = [object Object]`.

This PR fixes both of those errors but please correct me if I'm wrong and let me know if I can provide any further clarifications.

urbany:
This also happened to me and I was having a hard time understanding what happened. My case:
I have a list of items, and each item has a power-select. Every time a change a value (through power-select) i need to refresh the list, which means the current power-selects are destroyed before this close callback is called. When it's called since it is already destroyed an exception is thrown. I manage to work around the problem by delaying the action by a few MS, so the callback happens before I set my value, but it's a hack.

cibernox:
@urbany That actually seems a reasonable use-case and seems easy to reproduce. I'll try to get a failing test for that.

cibernox:
I was able to reproduce in the dummy app but no in testing (probably because of the frozen runloop). Either way, I've fixed in in ember-basic-dropdown instead, since the same thing could happen there: https://github.com/cibernox/ember-basic-dropdown/pull/158

cibernox:
Update to `beta.19`. This should be fixed.

urbany:
It's working great now @cibernox, thank you so much!

PR #644, opened 2016-08-22
fix 'Colletion' typo

barelyknown:


cibernox:
Good sight 👀 

Thanks!

PR #645, opened 2016-08-23
Dummy change to test that beta works

cibernox:


cibernox:
Finally Ember 2.8 is fixed!

PR #650, opened 2016-08-28
Fix #642: power-select tries to highlight an option from the wrong dropdown

### LINKED ISSUE #642: Cannot read property 'offsetTop' of null
Using Ember 2.7 and ember-power-select 1.0.0-beta.12, I have two basic selects set up like this:
```hbs
{{#power-select
  selected=vendor.country
  options=countries
  searchField='name'
  onchange=(action 'chooseCountry' vendor)
  as |country|
}}
  {{country.name}}
{{/power-select}}

{{#power-select
  selected=vendor.state
  options=states
  searchField='name'
  onchange=(action (mut vendor.state))
  as |state|
}}
  {{state.name}}
{{/power-select}}
```
I'll click on one, the options list opens, then, without closing it, I'll click on the other select, get the mentioned error in console while the other options list opens, and I'll get a `Cannot read property 'offsetTop' of null` error in console. After this, I can't close the second options list no matter what I try and basically both selects are unusable (also unable to select items).

I tried to set up a demo of this on `ember-twiddle`, but I'm not sure it supports addons quite yet.

jdurand:
fixes https://github.com/cibernox/ember-power-select/issues/642

### DIFF
--- a/addon/components/power-select.js
+++ b/addon/components/power-select.js
@@ -315,12 +315,14 @@ export default Component.extend({
       let index = indexOfOption(this.get('publicAPI').results, option);
       if (index === -1) { return; }
       let optionElement = optionsList.querySelectorAll('[data-option-index]').item(index);
-      let optionTopScroll = optionElement.offsetTop - optionsList.offsetTop;
-      let optionBottomScroll = optionTopScroll + optionElement.offsetHeight;
-      if (optionBottomScroll > optionsList.offsetHeight + optionsList.scrollTop) {
-        optionsList.scrollTop = optionBottomScroll - optionsList.offsetHeight;
-      } else if (optionTopScroll < optionsList.scrollTop) {
-        optionsList.scrollTop = optionTopScroll;
+      if (optionElement) {
+        let optionTopScroll = optionElement.offsetTop - optionsList.offsetTop;
+        let optionBottomScroll = optionTopScroll + optionElement.offsetHeight;
+        if (optionBottomScroll > optionsList.offsetHeight + optionsList.scrollTop) {
+          optionsList.scrollTop = optionBottomScroll - optionsList.offsetHeight;
+        } else if (optionTopScroll < optionsList.scrollTop) {
+          optionsList.scrollTop = optionTopScroll;
+        }
       }
     },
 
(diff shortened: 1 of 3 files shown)


cibernox:
Do you have a test to reproduce the failure? I am suspicious of having to use `run.later`.

If something is fixed with `run.later`, that is usually a symptom of something else.

jdurand:
It's easy to reproduce as described here https://github.com/cibernox/ember-power-select/issues/642, but I'll try to come up with a breaking test.

cibernox:
I tried yesterday, but before reading that last comment. I'm going to try to reproduce it again.

jdurand:
This should reproduce the bug, but the test doesn't fail... :-\
```
test('[BUGFIX] When the component opens, if another select with less options was previously opened', function(assert) {
  let done = assert.async();
  assert.expect(3);

  this.numbers = numbers;
  this.names = names;
  this.render(hbs`
    <div class="numbers">{{#power-select options=numbers selected="nine" onchange=(action (mut selected)) as |option|}}
      {{option}}
    {{/power-select}}</div>
    <div class="names">{{#power-select options=names selected="Marta" onchange=(action (mut selected)) as |option|}}
      {{option}}
    {{/power-select}}</div>
  `);

  clickTrigger('.names');
  run.later(() => {
    assert.equal($('.ember-power-select-option[aria-current="true"]').text().trim(), 'Marta');

    clickTrigger('.numbers');
    run.later(() => {
      assert.equal($('.ember-power-select-option[aria-current="true"]').text().trim(), 'nine');
      assert.ok($('.ember-power-select-options')[0].scrollTop > 0, 'The displayed list has scrolled');
      done();
    }, 120);
  }, 120);
});
```

cibernox:
Yeah, I'm trying myself too. Trying with an acceptance test too.

cibernox:
@jdurand I've fixed it in #651

The problem was that for scrolling I was taking the first list of option in the page, without taking into account that another select might be still opened, by example when it's being animated out.
This is a perfectly valid situation, so instead, the fix is ensure that I'm targeting the list of option **of this select**.

cibernox:
@jdurand Thanks for this! Pulling this string I found the exact problem!

jdurand:
That's the least I could do. Thanks for ember-power-select!

PR #657, opened 2016-09-05
add clear on touchstart

davidpett:
currently on iOS, the clear button opens the dropdown instead of clearing the selection. this adds the touch support


sebastianseilund:
This patch makes the clear button work on iOS in my app as well.

@davidpett could you make the tests pass? I think if you add `this.foo = 'one';` after `this.numbers = numbers;` then it'll work. The problem right now is that the clear button doesn't appear since nothing has been selected.

Then I think @cibernox can merge and we can get this in the next release.

Thanks!

cibernox:
There is a ESLint error, could you rebase and fix it?

addon/components/power-select/trigger.js
```diff
@@ -10,6 +10,9 @@ export default Component.extend({
     clear(e) {
       e.stopPropagation();
       this.get('select').actions.select(null);
+      if (e.type === 'touchstart') {
```
cibernox: Why do you return false here?

davidpett: to not bubble the event to the click which would open the dropdown

cibernox:
Thanks! I'll release a new version soon

PR #658, opened 2016-09-06
selected option should be not string

lifeart:
"Selected" option should be not string and some "array-like" object


cibernox:
I'm still catching up after holidays. What are you fixing here?

lifeart:
For multiselect case "selected" option can't be string, I guess. It should be kinda array

cibernox:
I see now. So at the moment if you pass a string the component breaks in some way and with your range is just going to ignore that string and return an empty array?

We might get the chance to give some helpful error message.

PR #671, opened 2016-09-17
Adjust copywriting on public-pages/addon/index

breyno127:
Some of the copy on this page felt off to me. This is just a proposed change to that language


cibernox:
Thanks!

PR #678, opened 2016-09-20
Fix keyboard navigation in the navigable-select demo

ef4:
Updating the customized `optionsComponent` in the demo so that keyboard navigation works again.


cibernox:
Yes, that demo has fallen way behind the current implementation. The failure in glimmer is expected

cibernox:
Thanks!

PR #679, opened 2016-09-20
Improved animation

ef4:
This switches from liquid-bind's `use=` to `rules=` so we can do declarative matching.

Also, both the existing code and my new version rely on equality checking that was not working with the latest code in power-select due to `toArray` on `options`. I patched the demo so that toArray is a no-op on our demo data, but it would be nice to skip the unnecessary copy in cases where options is already a true Array.

This PR builds off #678.


cibernox:
👍

PR #682, opened 2016-09-21
Open dropdown onInput after render to avoid deprecation warning

mdentremont:
- When triggering the power-select-multiple open when input is entered,
  only do so in the "afterRender" queue to avoid the
  "render-double-modify" deprecation warning

### DIFF
--- a/addon/components/power-select-multiple/trigger.js
+++ b/addon/components/power-select-multiple/trigger.js
@@ -80,7 +80,9 @@ export default Component.extend({
     onInput(e) {
       let action = this.get('onInput');
       if (action &&  action(e) === false) { return; }
-      this.get('select').actions.open(e);
+
+      const select = this.get('select');
+      scheduleOnce('afterRender', null, select.actions.open, e);
     },
 
     onKeydown(e) {


mdentremont:
Here are some steps to reproduce (with some of my guesses as to why this is occurring):

- Use a power-select-multiple which asynchronously searches (ie: via ajax calls)
- Click on the input field to show results
- Click one of the dropdown results, and at the same time hit a character on your keyboard
- This should simultaneously:
 - Select the clicked option, which will cause the dropdown to close, and the options to be refreshed
 - Trigger the dropdown to re-open, due to the keyboard input

cibernox:
I'll try to reproduce it first. It seems unintuitive to me to open the select in the after render, because that means that it won't open until the next frame.

cibernox:
@mdentremont I've been trying to reproduce but I can't se this error happening. I did a lot of click+types in every possible form. Are you in the latest version?

mdentremont:
@cibernox I'm trying hard to whip up a sample app for ya -- we have a lot of other stuff going on which surely could be the cause.

mdentremont:
FYI this is the deprecation warning, just in case you are curious:

```
DEPRECATION: You modified (-join-classes (-normalize-class "concatenatedTriggerClasses" concatenatedTriggerClasses) "ember-view" "ember-basic-dropdown-trigger" (-normalize-class "inPlaceClass" inPlaceClass activeClass=undefined inactiveClass=undefined) (-normalize-class "hPositionClass" hPositionClass activeClass=undefined inactiveClass=undefined) (-normalize-class "vPositionClass" vPositionClass activeClass=undefined inactiveClass=undefined)) twice in a single render. This was unreliable in Ember 1.x and will be removed in Ember 3.0 [deprecation id: ember-views.render-double-modify]
```

mdentremont:
@cibernox Doesn't seem to be an issue with this repo -- might be something to do with power-select-with-create (or how we might be incorrectly using it 😝

PR #695, opened 2016-09-30
Add selectedOptionClass

nicucampian:
- passing down to the selected li tag a specific selectedOptionClass

### DIFF
--- a/addon/templates/components/power-select.hbs
+++ b/addon/templates/components/power-select.hbs
@@ -73,6 +73,7 @@
     {{else}}
       {{#component optionsComponent
         class="ember-power-select-options"
+        selectedOptionClass=selectedOptionClass
         groupIndex=""
         loadingMessage=(readonly loadingMessage)
         id=(readonly optionsId)
--- a/addon/templates/components/power-select/options.hbs
+++ b/addon/templates/components/power-select/options.hbs
@@ -18,7 +18,7 @@
       {{/component}}
     </li>
   {{else}}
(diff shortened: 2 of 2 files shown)


nicucampian:
Hello, if more requirements for this to be merged in please let me know. Thanks

cibernox:
You already have an `aria-selected` attribute that is present on selected options. Your styles should target `.ember-power-select-option[aria-selected]`.

The reason why I want people to target that selector and not a class is because it ties together accessibility (A11y) good practices with styling. If programmers wants to create a customized version of EPS that reuses the existing themes, they have to add the right aria attributes.

It's me being a bit annoying for the sake of guiding people towards accesible components.

nicucampian:
if i target .ember-power-select-option[aria-selected] will change every selected options from every dropdown i have.

i managed it to work like :
dropdownClass="dropdown-3-class"

.dropdown-3-class > ul > li[aria-selected="true"]

and it works, but i want to make it more easy to add a styling. maybe i didn't understood right how it's working, and it is this solution

cibernox:
I'd use `.dropdown-3-class .ember-power-select-option[aria-selected]`. 

As I mentioned, the reason is that I want to force people to target `aria-*` attributes for styling instead of classes to enforce good a11y.

The PR has nothing wrong technically, it's just a philosophical decision of the component.

Thanks for dedicating the time for this anyway.

nicucampian:
Oke, i will comply due to [aria] . Thanks anyway

PR #703, opened 2016-10-01
Add support for cancellables

### LINKED ISSUE #693: Add support for `cancel`ables in search
As of today, EPS is already `then`able aware. That means that some actions like `search` expect the value returned from user-land to to have a `then` method that, eventually, resolves to a collection.

At the moment, when a search is fired before the previous one is resolved, EPS is smart enough to just care about the most-recent promise, so even when those promises resolve, the component ignores it's result.

However, ignoring the result doesn't prevent the job from happening, which is sub-optimal. Ideally EPS should cancel work that is going to be ignored when possible.

There is no such thing as cancellable promises (yet), but in ember we have `ember-concurrency`'s tasks, that have a thenable interface and also have a `cancel` method.

EPS must detect the presence of a `cancel` method and use it where possible.

cibernox:
With ember-concurrency gaining popularity every day and cancellable promises
around the corner (sic) it's time to start being good citizens and cancel all
work that can be cancelled when the component is no longer interested in it.

There is a couple situations where this can happen:

If the `search` action returns a `thenable` object that responds to the
`.cancel()` method, firing another search before the first one is completed will
cancel the previous. This was possible before by just adding `.restartable()`
to your e-c task, but now it will work with any cancellable object.

Another situation where this could happen is when, after firing a search that
returns can cancellable-thenable, and before that search completes, the searchbox
gets cleared. In EPS, clearing the search doesn't fire another search, so this
situation was not easy to detect and cancel properly with a `.restartable()` task.
Now, EPS detects this situation and cancels any pending search when the searchbox
is cleared.

Closes #693

### DIFF
--- a/addon/components/power-select.js
+++ b/addon/components/power-select.js
@@ -117,6 +117,7 @@ export default Component.extend({
     this.activeSelectedPromise = this.activeOptionsPromise = null;
     this._removeObserversInOptions();
     this._removeObserversInSelected();
+    this._cancelActiveSearch();
     cancel(this.expirableSearchDebounceId);
   },
 
@@ -435,6 +436,7 @@ export default Component.extend({
 
   _resetSearch() {
     let results = this.get('publicAPI').options;
+    this._cancelActiveSearch();
     this.updateState({
       results,
       searchText: '',
@@ -458,6 +460,7 @@ export default Component.extend({
     if (!search) {
       publicAPI = this.updateState({ lastSearchedText: term });
     } else if (search.then) {
+      this._cancelActiveSearch();
       publicAPI = this.updateState({ loading: true, _activeSearch: search });
       search.then((results) => {
         if (this.get('isDestroyed')) {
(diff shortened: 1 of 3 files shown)


addon/components/power-select.js
```diff
@@ -570,6 +573,13 @@ export default Component.extend({
     }
   },
 
+  _cancelActiveSearch() {
+    let publicAPI = this.get('publicAPI');
+    if (publicAPI._activeSearch && typeof publicAPI._activeSearch.cancel === 'function') {
+      publicAPI._activeSearch.cancel();
+    }
+  },
+
```
sukima: FYI, Ember has a `tryInvoke` that does just this without the boilerplate.

```js
let activeSearch = this.get('publicAPI._activeSearch');
Ember.tryInvoke(activeSearch, 'cancel');
```

http://emberjs.com/api/#method_tryInvoke

Also I just checked and `tryInvoke` is 'null' / 'undefined' safe.

PR #709, opened 2016-10-06
Add `onblur` action for symmetry with `oninput`

### LINKED ISSUE #708: add an `onblur` action, for symmetry with `onfocus`
Investigation of how `onfocus` works, so `onblur` behaves consistently.

- It fires when the trigger of a single select is focused
- If ALSO fires when the searchbox of the single select is focused, even if the trigger already had the focus.
- It fires when the trigger of a multiple-select gets focused, but ...
- it DOES NOT fire when the input inside gets focused. And when there is an input, the trigger never gets focused.

What I think is the right behaviour (balancing usability and versatility):

A select may have many components inside that can gain or loose the focus, but it's not feasible to have different events for each one of them (`triggerFocus` vs `searchboxFocus` vs `whoknowsFocus`). 

So, I think that the simplest a more consistent solution is to, as it works now, fire it from any element that might gain the focus. Since the second argument of the action is the event itself, it can be used to disambiguate like this:

```js
handleFocus(select, e) {
  if (e.target.classList.includes('.ember-power-select-searcbox')) {
    // do something that I only do when the searchbox is focused.
  }
}
```

This also means than, by example, in the case of a component that opens on focus, like this:

```hbs
{{#power-select ... onfocus=(action "handleFocus") onblur=(action "handleBlur")}}
```

Given that the trigger is focused when it's clicked, but then it opens and the searchbox is focused as soon as the component opens, three events will fire in quick succession: `onfocus` (in the trigger), `onblur` (in the trigger) and `onfocus` (this time in the searchbox).

If the actions performed by `onfocus`/`onblur` enter in conflict, the user should take care himself of delaying/coalescing actions if needed.

cibernox:
Closes #708

### DIFF
--- a/addon/components/power-select.js
+++ b/addon/components/power-select.js
@@ -341,6 +341,22 @@ export default Component.extend({
       }
     },
 
+    onTriggerBlur(_, event) {
+      this.send('deactivate');
+      let action = this.get('onblur');
+      if (action) {
+        action(this.get('publicAPI'), event);
+      }
+    },
+
+    onBlur(event) {
+      this.send('deactivate');
+      let action = this.get('onblur');
+      if (action) {
+        action(this.get('publicAPI'), event);
+      }
+    },
+
     activate() {
       this.updateState({ isActive: true });
     },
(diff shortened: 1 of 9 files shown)


miguelcobain:
@pauln maybe ember-paper can now leverage core functionality?

cibernox:
I haven't published a new version yet, but I will over the weekend.

PR #721, opened 2016-10-28
lowercase string data type in api-reference.md

etseidler:
for consistency with how that data type is listed for all the other API options


cibernox:
Good eye

PR #722, opened 2016-10-28
add ability to move list of selected items to another element…

etseidler:
… via css selector

I recently did some custom styling on the power-select-multiple where I wanted to move the list of selected items to a DOM element outside of the component. I thought it might be helpful as an option for others who may also be doing custom styling.

Let me know what you think.

### DIFF
--- a/addon/components/power-select-multiple.js
+++ b/addon/components/power-select-multiple.js
@@ -2,12 +2,14 @@ import Component from 'ember-component';
 import computed from 'ember-computed';
 import layout from '../templates/components/power-select-multiple';
 import fallbackIfUndefined from '../utils/computed-fallback-if-undefined';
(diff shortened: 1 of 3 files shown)


cibernox:
I wouldn't do this this way. It doesn't work really beyond initial render. I doesn't update. I doesn't work in fastboot either.

The correct way of doing this would be to just render things normally wherever you want and pass a custom triggerComponent with a different template that doesn't render the selected items.

```hbs
{{#power-select-multiple options=dogs selected=myDogs triggerComponent="custom-component-that-renders-nothing" as |dog|}}
  {{dog.name}}
{{/power-select}}

<ul class="somewhere-else">
  {{#each myDogs as |dog|}}
    {{dog.name}}
  {{/each}}
</ul>

PR #724, opened 2016-11-01
Fix error constructing Event in Internet Explorer 11

dwickern:
Tests which use the `clickTrigger` helper fail in IE11.


cibernox:
Interesting. Thanks!

PR #728, opened 2016-11-12
Fix debounced search with ember-concurrency documentation

acburdine:
As seen [here](http://ember-concurrency.com/#/docs/task-concurrency) the default option for ember-concurrency tasks is to run concurrently, which doesn't make much sense for a debounced task 😉 

I've updated both the code snippet and the actual controller to have the `restartable` modifier, which should enable true debouncing.

### DIFF
--- a/tests/dummy/app/controllers/public-pages/cookbook/debounce-searches.js
+++ b/tests/dummy/app/controllers/public-pages/cookbook/debounce-searches.js
@@ -17,7 +17,7 @@ export default Ember.Controller.extend({
     yield timeout(600);
     let url = `https://api.github.com/search/repositories?q=${term}`;
     return this.get('ajax').request(url).then((json) => json.items);
-  }),
+  }).restartable(),
 
   _performSearch(term, resolve, reject) {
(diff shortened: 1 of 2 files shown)


acburdine:
@cibernox I edited this on Github, so I wasn't able to squash the two commits into one :) If you want, I can pull this branch later and squash the commits together, or otherwise just squash it on merge.

cibernox:
Actually, that example is on purpose. Ember Power Select task aware **task-aware**. That means that if you return to the search function a promise-like object that has a `.cancel()` method, EPS will call it automatically.
It's even smarter than just adding `.restartable()`, because it also cancels ongoing tasks when the component is closed or destroyed.

So the point is that no matter if the task is concurrent, `.restartable()`, `.drop()`, `.enqueue()` or it's not a task at all (thinking about the future cancellable promises), all are going to be cancelled.

I'd accept however a PR adding a comment explaining that this is in purpose.

Thanks!

acburdine:
@cibernox Ah, thanks for the clarification!

PR #732, opened 2016-11-15
Search Input Focus Bug

designblooz:
If I'm using ember-power-select-typeahead with ember-power-select, it focuses to the first element found for `.ember-power-select-search-input` not the direct children search bar input. 

This will bind the querySelector to the power-select component, not the whole dom.

### DIFF
--- a/addon/components/power-select/before-options.js
+++ b/addon/components/power-select/before-options.js
@@ -1,3 +1,4 @@
+import Ember from 'ember';
 import Component from 'ember-component';
 import { scheduleOnce } from 'ember-runloop';
 import layout from '../../templates/components/power-select/before-options';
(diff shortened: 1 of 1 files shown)


designblooz:
Seems like dependency update failure.

cibernox:
This is a failure indeed, but I think that there is a slightly better way of fixing this (without using `Ember.get(this, 'parentView.element')`).

I'll check it tomorrow.

dwickern:
Temporary workaround if anyone needs
```js
import Ember from 'ember';
import BeforeOptions from 'ember-power-select/components/power-select/before-options';

BeforeOptions.reopen({
  focusInput() {
    const parentElement = Ember.get(this, 'parentView.element');

    if (parentElement) {
      this.input = parentElement.querySelector('.ember-power-select-search-input');
    }
    if (this.input) {
      Ember.run.scheduleOnce('afterRender', this.input, 'focus');
    }
  }
});
```

cibernox:
@dwickern Can you update to the latest version? I think that this workaround is not needed anymore in 1.7.2

cibernox:
Closing very old PR

PR #742, opened 2016-11-29
Pass disabled state to the trigger component

dwickern:
ember-power-select-typeahead needs to know the disabled state in order to disable its `<input>` element


cibernox:
If i'm right, the publicAPI of the component, named `select` is available in all components and already has the disabled property you want: http://www.ember-power-select.com/docs/api-reference

PR #763, opened 2016-12-14
Document using POJOs as selected

orf:
Is this OK?

### DIFF
--- a/tests/dummy/app/templates/public-pages/docs/how-to-use-it.hbs
+++ b/tests/dummy/app/templates/public-pages/docs/how-to-use-it.hbs
@@ -57,6 +57,15 @@
 </ul>
 
 <p>
+  Using a POJO as the `selected` value is not supported by design. If you want to use a list of POJOs as the options
+  then you can use `ember-composable-helpers` to implement this:
+</p>
+
+{{#code-sample hbs="how-to-use-it-3.hbs" as |partialName|}}
+  {{partial partialName}}
+{{/code-sample}}
+
+<p>
   This is the very basic usage of this component, but so far is pretty useless. Since there are no
   two way bindings and the <code>onchange</code> action is empty <strong>nothing happens when you select something.</strong>
 </p>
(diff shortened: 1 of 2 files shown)


omairvaiyani:
@orf Looks like the test failed because 'find-by' was not found as a helper. 

You need to add [ember-composable-helpers](https://github.com/DockYard/ember-composable-helpers) to this repos dependencies. Ideally limiting to the tree to 'find-by' like below.

```javascript
// ember-cli-build.js
var app = new EmberApp(defaults, {
    ...
    'ember-composable-helpers': {
      only: ['find-by']
    }
    ...
  });
```

cibernox:
I think that when this PR was created master was red but now it should pass the tests after a rebase

orf:
Oh dear. I rebased it on master and now it's a tad fucked it seems!

orf:
So it's failing because `unit/utils/group-utils-test.js: should pass ESLint` which I haven't touched. Is it OK to merge anyway?

cibernox:
I don't see that ESlint error that you mention. However I'm going to hold this a bit more until I implement #773, since can change the narrative or add another option to the explanation.

cibernox:
I think I can close this since this part of the docs have already been improved.

Caltor:
@cibernox The docs still don't say how to handle this using a computed property or `find-by`.

cibernox:
@Caltor where would you add an example for that? Maybe the "How to use" section?
If you have time to make a PR improving the docs I think that would be valuable. Often library authors have blindspots because we know the library too well.

Caltor:
@cibernox Yeah I would suggest the "How to use" section. If you're happy for me to mess with the docs I'll give it a go. It will be only my second PR tho so please be gentle with me. ;)

cibernox:
Go ahead! The docs are the dummy app in this very repo. More often that not, people with less knowledge of the library are the one who write the better docs, as they explain things more thoroughly.

Caltor:
@cibernox OK I've created https://github.com/cibernox/ember-power-select/pull/1177. Hopefully I've done it right but please let me know if I've messed up.

Caltor:
@cibernox Looks like the build is failing but it has been failing for the last 28 days so I don't think it is anything I have done. See https://travis-ci.org/cibernox/ember-power-select/builds

Caltor:
@cibernox I've fixed the build issue on a separate PR for you in case you reject the doc changes PR (which also incorporates the fix) so you can take your pick.

PR #764, opened 2016-12-16
hide search message if options are loading

miguelcobain:


cibernox:
Thanks! I'd release shortly

PR #767, opened 2016-12-23
Update ember-cli-sass v6.0.0

xomaczar:
Support node-sass v4.1.0 Alphine linux


cibernox:
@xomaczar if you can rebase, I think that this should be green now

xomaczar:
Good to go!

PR #781, opened 2017-01-13
Update node version in package.json

simonihmig:
Since you use ES6 in node-land since #776 this should be reflected in `package.json`


cibernox:
Thank you sir 🎩

PR #790, opened 2017-01-19
add extra hash to options component

balajimanoharan:
add extra hash to options component


cibernox:
Thanks! I'll release a new version in a moment.

PR #801, opened 2017-01-30
Fix app.import bug

ssriera:
## Problem
The build (in EPSWFB) was failing due to `app.import is not a function` in index.js of EPS.

## Approach
Add an extra layer that checks if the app is located in `app.parent.app`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,11 +1,19 @@
 /* eslint-env node */
 'use strict';
 
+function lookupApp(appOrAddon) {
+  if (appOrAddon.app) {
+    return appOrAddon.app || appOrAddon;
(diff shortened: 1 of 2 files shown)


ssriera:
I wasn't getting the correct error message so I had to clear the npm cache & remove the `yarn.lock`. I yarned and that's why that file is updated- I know that shouldn't be included in this PR _but_ tests passed so I figured why not.

nathanhammond:
@cibernox @ssriera Use `this.import` in CLI 2.7+ and it will always do the right thing.

If you need to do this and support CLI pre-2.7, polyfill `this.import` and `_findHost` so that this works inside of engines.

PR #808, opened 2017-02-03
Add identification for the trigger display yield

monovertex:
This PR adds information in the yielded block whether the block is rendered in the trigger or in the dropdown, to allow for different logic in those places.

My use case: I implemented a lazy rendering component and I want to wrap the dropdown options in it, but I don't want the trigger block to have the lazy loading behavior as well. I feel like this could be used in many other situations.

### DIFF
--- a/addon/templates/components/power-select.hbs
+++ b/addon/templates/components/power-select.hbs
@@ -38,7 +38,7 @@
       select=(readonly publicAPI)
       selectedItemComponent=(readonly selectedItemComponent)
       as |opt term|}}
-      {{yield opt term}}
+      {{yield opt term true}}
     {{/component}}
   {{/dropdown.trigger}}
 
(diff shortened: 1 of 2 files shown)


cibernox:
The say I prefer people to make the trigger and the options diverge is by using the `selectedItemComponent` (see http://www.ember-power-select.com/docs/the-trigger)

While slightly more inconvenient (you have to create a template-only component) it is more powerful and generally faster, because adding and `if/else` in the block forces you to evaluate that if else N number of times, one per each item of the collection. With 10 items doens't matter, with 150 it does.

monovertex:
The problem with this approach is that it requires multiple `selectedItemComponents` if different EPS instances have different displays. This overhead becomes even bigger if your trigger display and option display are really similar, because then you probably need a new component to use in both the `selectedItemComponent` and the yielded block of the EPS, to avoid duplicate code.

Performance is, however, more important I think. I'll think about it, maybe we can find a better way. I feel like the scenario described by me earlier can become pretty performance-intensive as well, because component instantiation is also expensive, same as evaluating `if/else` constructs.

PR #818, opened 2017-02-15
Allow child to override parent options

lazybensch:
What this PR does is it enables you to override "parent options" in your custom components like this:
```
{{#power-select
   options=countries
     selected=country
     searchEnabled=false
     beforeOptionsComponent=(component "power-select/before-options" searchEnabled=true)
     onchange=(action (mut foo))
     as |country|}}

  {{country.name}}
{{/power-select}}
```

`searchEnabled` is set to `false` in the **power-select** component but explicitly set to `true` inside the custom **beforeOptionsComponent** (this is currently not possible).

When you don't specify an option explicitly in the component helper it will default to the option specified in **power-select** as it's currently the case in master.

----

In this PR I only enabled this behaviour for the `searchEnabled` option and the `beforeOptionComponent`. However i wrote a [helper](https://github.com/cibernox/ember-power-select/compare/master...lazybensch:allow-child-to-override-parent-options?expand=1#diff-a06055fe46f893c508b99880c5d693abR53) that can be reused for any option that you feel should be overridable.

### DIFF
--- a/addon/components/power-select/before-options.js
+++ b/addon/components/power-select/before-options.js
@@ -1,4 +1,5 @@
 import Component from 'ember-component';
+import { or } from 'ember-computed';
 import { scheduleOnce } from 'ember-runloop';
 import layout from '../../templates/components/power-select/before-options';
 
@@ -6,6 +7,8 @@ export default Component.extend({
   tagName: '',
   layout,
   autofocus: true,
+
+  internalSearchEnabled: or('searchEnabled', 'externalSearchEnabled'),
 
   // Lifecycle hooks
   didInsertElement() {
@@ -18,7 +21,7 @@ export default Component.extend({
 
   willDestroyElement() {
     this._super(...arguments);
-    if (this.get('searchEnabled')) {
+    if (this.get('internalSearchEnabled')) {
       scheduleOnce('actions', this, this.get('select').actions.search, '');
     }
   },
--- a/addon/components/power-select/before-options.js
+++ b/addon/components/power-select/before-options.js
@@ -1,5 +1,4 @@
 import Component from 'ember-component';
-import { or } from 'ember-computed';
 import { scheduleOnce } from 'ember-runloop';
 import layout from '../../templates/components/power-select/before-options';
 
@@ -7,8 +6,6 @@ export default Component.extend({
   tagName: '',
   layout,
   autofocus: true,
-
-  internalSearchEnabled: or('searchEnabled', 'externalSearchEnabled'),
 
   // Lifecycle hooks
   didInsertElement() {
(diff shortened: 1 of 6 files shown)


cibernox:
I find this weird. What do you want to achieve with this? Why not set it to true?

I'd suggest to pass your own custom component instead if you want to have a searchbox, even if you just copy-paste/reexport the original implementation of the one provided by this component.

lazybensch:
I think there are a lot of places where the upside to this approach would be more clear, for example:

currently the `triggerComponent` and `beforeOptionsComponent` share a lot of the same options like searchEnabled, onFocus, onInput... and I could see how users would want a way to set up different events without having to write their custom component. (Its btw not as easy as writing your own custom beforeOptionsComponent because power-select would still feed it with its options, you would have to write a custom ember-power-select component that inherits the original one and that feels quite cumbersome to me).

Actually. **placeholder**! the placeholder option is a brilliant example. You seem to agree that the triggerComponent and the beforeOptionsComponent should be configurable to have different placeholders. In the current implementation you achieve this by:

```
{{#power-select
   options=countries
     selected=country
     placeholder="foo"
     searchPlaceholder="bar"
     beforeOptionsComponent=(component "power-select/before-options" autofocus=true)
     onchange=(action (mut foo))
     as |country|}}

  {{country.name}}
{{/power-select}}
```

With the helper I introduced in this PR you could easily make the placeholder API to also accept:


```
{{#power-select
   options=countries
     selected=country
     placeholder="foo"
     beforeOptionsComponent=(component "power-select/before-options" placeholder="bar" autofocus=true)
     onchange=(action (mut foo))
     as |country|}}

  {{country.name}}
{{/power-select}}
```

Which to me, looks way cleaner. Writing `placeholder="bar"` as an option on the component it belongs to instead of passing it to the parent component and have it pass it down for you.

cibernox:
I created the `extra` property as an scape valve when you want to pass options beyond what the public API is designed for.

I'm not entirely sure wheter this is using only public API, since the way contextual components are designed is that the last level wins.

I believe that the **right** solution, although admitely more involved is to step away from the current API and expose a lower-level public API based in contextual components, much like ember-power-calendar or ember-power-datepicker.

With that, users would have control because **they** would be the one calling the component the last time, so their options would win.

Example API:

```hbs
{{#power-select-basic someGlobalOptions as |select|}}
  {{select.trigger}}
  {{#select.content}}
    {{select.beforeOptions placeholder="placeholder"}} {{! you win}}
    {{#select.options as |opt|}}
      {{opt}}
    {{/select.options}}
  {{/select.content}}
{{/power-select}}
```

Obviously the current API would continue to exist as a layer above this lower-level abstraction for backwards compatibility reasons (and because this is too verbose for the 99% of the time you just want a select)

PR #821, opened 2017-02-17
don't use ember-truth-helpers in addon directly

### LINKED ISSUE #810: remove ember-truth-helpers from dependencies
is it really necessary to bring them in as a dependency? Would you consider moving them to devDependencies for the dummy app, and I could submit a PR that replaces all usage of truth helpers in the addon with computed properties?

akatov:
fixes #810

### DIFF
--- a/addon/components/power-select-multiple/trigger.js
+++ b/addon/components/power-select-multiple/trigger.js
@@ -2,7 +2,7 @@ import Ember from 'ember';
 import Component from 'ember-component';
 import layout from '../../templates/components/power-select-multiple/trigger';
 import get from 'ember-metal/get';
-import computed from 'ember-computed';
+import computed, { and, not } from 'ember-computed';
 import service from 'ember-service/inject';
 import { scheduleOnce } from 'ember-runloop';
 import { assert } from 'ember-metal/utils';
@@ -20,6 +20,9 @@ export default Component.extend({
   layout,
   textMeasurer: service(),
   _lastIsOpen: false,
+
+  _searchDisabled: not('searchEnabled'),
+  _placeholderAndSearchDisabled: and('placeholder', '_searchDisabled'),
 
   // Lifecycle hooks
   didInsertElement() {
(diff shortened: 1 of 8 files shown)


akatov:
@cibernox what do you think?

cibernox:
I thought that it was a WIP. Those examples look good, but there is still some usages. The tricky ones. Example: https://github.com/cibernox/ember-power-select/blob/master/addon/templates/components/power-select/options.hbs#L24

akatov:
thanks! I totally missed that one. Will send you an update shortly

akatov:
It's a bit more complex to use computed properties, since some other parts of the `publicAPI` get affected by the change of what is passed to the options component. In particular the highlighted property.
I'd be happy to look into doing this properly in a separate PR, but for now I just wanted to finish off this one...

cibernox:
Yeah, that is what I though. I'm unsure if I should merge this, because if it doesn't fully achieve the final goal (remove a runtime dependency) there is no benefits and a certain (albeit small) amount of churn for people who had created their own components and copy/pasted the original templates.

Let's think on a plan to remove it entirely.

akatov:
well, with the final commit the goal of removing the runtime dependency is achieved (by replacing the eq helper with a custom helper). Moreover, this should still work with the old templates, since the data didn't change.
Then again I'm not sure whether this would warrant a breaking change nevertheless because if someone was using e-p-s and was using truth helpers without explicitly specifying them as a dependency, they would have to now require them explicitly.
In the end, this PR probably doesn't provide enough benefit overall, especially because it makes things more complex in the templates, which as you said are part of the public API, so maybe it is better to hold off until we change the actual data and can remove all the custom template helpers (which would certainly warrant a major version bump).

cibernox:
After some thought I decided to continue using the helpers. It is the 6th most used addon in the ecosystem (it will be present it most apps anyway) and it's height footprint is so small that I think it's not worth it.

If I remove a runtime dependency, it would be ember-concurrency as it's not so small.

PR #823, opened 2017-02-19
Allow unselection in select fn

### LINKED ISSUE #822: Allow onchange to optionally clear selected value when selecting the selected value?
I had anticipated that selecting an item would allow for a "deselection".

It appears, however, that the addon explicitly prevents this behavior:

```js
select(selected /* , e */) {
  let publicAPI = this.get('publicAPI');
  if (!isEqual(publicAPI.selected, selected)) {
    this.get('onchange')(selected, publicAPI);
  }
},
```

I would love to have `onchange` (i.e. `select`) only compare with `isEqual` by default, but be configurable to pass the action upwards, perhaps with an `allowUnselect` or similar.

The documentation around `onchange` is confusing, too, because it looks as though this should be possible:

> The function to be invoked when the user selects or **unselects** an option

Can you clarify if I'm misunderstanding or misusing the addon? And in the event unselecting is not the desired behavior, could we propose allowing this instead?

Thank you!

JoshSmith:
**EDIT**

This PR now documents the use of `buildSelection` and reorders the API reference. I'm happy to circle back later and create an example of this, but for now need to return to some other work.

Will open a follow-up issue to do so once this is merged.

<hr />

Closes #822.

Just an initial PR in the event that we want to allow unselection via a configuration option. Obviously open to any kind of feedback and understand that, if approved, will require some documentation.

### DIFF
--- a/tests/dummy/app/templates/public-pages/docs/api-reference.hbs
+++ b/tests/dummy/app/templates/public-pages/docs/api-reference.hbs
@@ -12,6 +12,16 @@
   </thead>
   <tbody>
     <tr>
+      <td>afterOptionsComponent</td>
+      <td><code>string/contextua-component</code></td>
+      <td>The component rendered after the list of options. It is empty by default in both single and multiple selects</td>
+    </tr>
+    <tr>
+      <td>allowClear</td>
+      <td><code>boolean</code></td>
+      <td>When truthy, single selects allow to nullify the selection</td>
+    </tr>
+    <tr>
       <td>ariaDescribedBy</td>
       <td><code>string</code></td>
       <td>Sets <code>aria-describedby</code> on the component</td>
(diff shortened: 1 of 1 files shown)


cibernox:
The addon already has an `allowClear` option for that purpose. In the default implementation it draws a small `×` in the trigger to unselect the selected option. You can see an example at the bottom of [this section](http://www.ember-power-select.com/docs/the-trigger)

This should cover what you want.

JoshSmith:
@cibernox I'm specifically hoping for the ability to have a deselection from the results. In the sidebar of this PR, for example, you can play with the assignment select UI. This is a similar outcome.

JoshSmith:
For example, the test user here is the current selection and I want the `onchange` to be fireable with this selection. I would be equally happy to have an `onselect` or something similar exposed.

cibernox:
I see. I still think that this is beyond what a "normal" select is so it shouldn't be in the core of the component. It's rather custom.

What I'd do in your feet to get that is to pass a custom `optionsComponent`. 

Steps:
- Create the component. Make it 
```
import OptsComponent from 'ember-power-select/components/options

export default OptsComponent.extend({
  actions {
    clear() {
      e.stopPropagation();
      this.get('select').actions.select(null);     
    }
  }
});
```
- Copy the [original template](https://github.com/cibernox/ember-power-select/blob/master/addon/templates/components/power-select/options.hbs) and paste it in the template of your new component. Simplify parts if you don't care about  them (like groups).
- Add the `×` button there. When clicked it must the "clear" action defined there.


Done (Or so I think)

JoshSmith:
I'm sorry if I'm confused here, but this seems to be a fairly common pattern in select UIs. There are multiple examples in the sidebar here on GitHub. Trello has several examples, as does Asana.

Definitely appreciate the thoroughness of direction for an alternative, but does an `onselect` event that bubbles seem outside the scope of what you'd intended here?

JoshSmith:
I'd actually probably be more inclined towards an `onselect` vs a configurable `onchange`, as the latter does feel quirky in the sense that it is not a _change_.

JoshSmith:
Also, if this is more the domain of an addon, I could see that, although it might end up being a silly small addon.

cibernox:
I've checked this again with caffeine.

Ok, there is a semi-public API of doing this. The `buildSelection` option. I don't think I've ever documented this. Try this:

```hbs
{{#power-select buildSelection=(action "nullIfSelected") selected=selected options=options ...}}
```

```js
actions: {
  nullIfSelected(option, select) {
    if (Ember.isEqual(select.selected)) {
      return null;
    }
    return option;
  }
}
```

If this works for you, you can implement it yourself by extending the component like explained in [this section](http://www.ember-power-select.com/cookbook)

```js
// app/components/power-select.js
import EmberPowerSelect from 'ember-power-select/components/power-select';

export default EmberPowerSelect.extend({
  buildSelection(option, select) {
    if (this.get('allowUnselect') && Ember.isEqual(option, select.selected)) {
      return null;
    } else {
      return this._super(...arguments);
    }
  }
});
```

cibernox:
I never documented this option but it's part of the public API. This function is the one that receives one option and returns the value with which the `onchange` action is going to be invoked.
In your case, if your selected option is the same that is already selected, you want to select null.

This is already how the multiple select makes to select/unselect options (and transform it into an array). Check: https://github.com/cibernox/ember-power-select/blob/master/addon/components/power-select-multiple.js#L85-L101 and https://github.com/cibernox/ember-power-select/blob/master/addon/templates/components/power-select-multiple.hbs#L10

If this works for you, I'd suggest you to use this very PR to edit the docs and make it public. You could even add an example of the docs of how to use it.

JoshSmith:
Awesome thanks for looking into this. I was wondering if there was a way that the multiple select was handling this to allow deselection, so glad to hear that intuition was directionally accurate. If I do get this working, very happy to update docs with an example and expose publicly.

cibernox:
Thanks! I'll update the docs now.

PR #826, opened 2017-02-24
allow to specify ebd content tag name

miguelcobain:
in order for us to be able to take advantage of https://github.com/cibernox/ember-basic-dropdown/pull/236


miguelcobain:
Probably a not needed `fallbackIfUndefined`, but you never know.

PR #827, opened 2017-02-27
Disable trigger when option is disabled

tschoartschi:
I added behavior to `templates/components/power-select-multiple/trigger.hbs` so that the remove-button is hidden if an option is disabled. In our app this happens within the following use case: 

There are options which are only addable by an admin. The regular user should see the options but should not be able to remove them. 

In our UI this looks something like this:
image

The green category are the "admin categories". They should not be deletable by the regular user. Therefore the disabled flag is set on them.

@cibernox let me know what you think. Would be great to have a way to "disable" single options. Maybe my solution is sufficient 😉

### DIFF
--- a/addon/templates/components/power-select-multiple/trigger.hbs
+++ b/addon/templates/components/power-select-multiple/trigger.hbs
@@ -1,7 +1,7 @@
 <ul id="ember-power-select-multiple-options-{{select.uniqueId}}" class="ember-power-select-multiple-options">
   {{#each select.selected as |opt idx|}}
-    <li class="ember-power-select-multiple-option">
-      {{#unless (or select.disabled opt.disabled)}}
+    <li class="ember-power-select-multiple-option {{if opt.disabled 'ember-power-select-multiple-option--disabled' }}">
+      {{#unless select.disabled }}
         <span role="button"
           aria-label="remove element"
           class="ember-power-select-multiple-remove-btn"
--- a/addon/templates/components/power-select-multiple/trigger.hbs
+++ b/addon/templates/components/power-select-multiple/trigger.hbs
@@ -1,7 +1,7 @@
 <ul id="ember-power-select-multiple-options-{{select.uniqueId}}" class="ember-power-select-multiple-options">
   {{#each select.selected as |opt idx|}}
     <li class="ember-power-select-multiple-option">
-      {{#unless select.disabled}}
+      {{#unless (or select.disabled opt.disabled)}}
         <span role="button"
           aria-label="remove element"
           class="ember-power-select-multiple-remove-btn"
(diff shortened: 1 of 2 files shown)


cibernox:
This is an interesting question. I'm not sure what the right behaviour is. 

It seems weird that a disabled option is selected in the first place, so I'm unsure if this behaviour should be baked in the default component or it should be up to you to do it in your selects (basically do this in your own custom trigger component).

Can you check what other libraries do in this situation? 

I think that your approach makes sense (although the PR would require tests)

tschoartschi:
I am also not sure what the right behaviour should be. In our case it makes sense but I'm also not sure if it makes sense for a generic solution. I'll try to find some other libraries to compare. Writing my own trigger component would also be possible but I really just want to hide the delete-button and use all the other functionality of your trigger. Not sure how to achieve this in a clever way.

If we decide that the behaviour of this pull request makes sense I'll write tests 😉 should I put it into `integration/components/power-select/multiple-test.js`?

cibernox:
If you want to make this behaviour the default for your app you basically have to do the same thing you did, but in your app.

```js
// in app/components/power-select-multiple/trigger.js
import MultipleTrigger from 'ember-power-select/components/power-select-multiple/trigger';
export default MultipleTrigger.extend(); // This will replace the default trigger with your version
```

and create `app/templates/components/power-select-multiple/trigger.hbs` that is a copy-paste of the original template with the change you made in this PR.

tschoartschi:
> that is a copy-paste of the original template

And this is the thing I wanted to avoid because then I have to track your `app/templates/components/power-select-multiple/trigger.hbs` for changes. Not sure how often it will change or if it changes at all. But I use everything what `trigger.hbs` provides, from CSS to functionality. 
And if I update `ember-power-select` in let's say 6 or 12 month I'm sure I'll forget to check `trigger.hbs` for changes 😆

cibernox:
Understandable. Although since templates are public-ish template it I change them it should be a major version bump.

Either way, if you can find some prior art on this subject that would be great. So some extent it makes sense that disabled options can't be interated with, for adding it nor for removing it.

cibernox:
I had another idea. Something I would approve immediately without looking for prior art is adding a class `ember-power-select-multiple-option--disabled` to those options that are disabled.

This is not a change in behaviour and you can use CSS to make the `x` button be `display: none`.

tschoartschi:
Would be fine for me too. I'll change my pull request. Just give me a second 😉

tschoartschi:
Travis failed... don't know why. @cibernox do you have any idea were I should start to fix this problem?

cibernox:
Im 90% it's a random failure. Let me rerun that example

tschoartschi:
@cibernox I'm happy that I didn't break something. Now it seems fine 😄

cibernox:
Released in 1.5.0-beta.2

Thanks!

PR #828, opened 2017-02-27
Update EBD with breaking change that never surfaced EPS's api

cibernox:
People using EPS will not have any breaking change


miguelcobain:
shouldn't we take this chance to forward the positioning option to ebd?

cibernox:
Sure, I'm just doing it on two steps

PR #852, opened 2017-03-12
Update dependencies for more babel@6 testing.

rwjblue:
PR against babel@6 branch, to see how CI fairs with the updated deps.

`ember test` now passes though, so we are heading in the right direction...


rwjblue:
Looks like the same jobs are failing here as on master (haven't dug into what is causing master to fail though).

cibernox:
Seems unrelated to me too. I'll merge it and check what is the cause (if it's not spurious) in the other branch.

cibernox:
@rwjblue The error seems to be a regression introduced in 2.11.3. More info here: https://github.com/emberjs/ember.js/pull/14898

cibernox:
In the meanwhile I'm going to comment that test.

PR #864, opened 2017-03-21
update `contentId` during async tests

noslouch:
I've got a test that's failing b/c the power select is only rendered when the promise is fulfilled. Due to the async op, `.ember-power-select-trigger` isn't rendered, so `$trigger` is defined in the fallback clause. The element that's defined for `$trigger` is outside the scope of the `power-select` component, and so has no `aria-owns` attribute.

This patch simply runs through the "find a trigger" logic once again in the `andThen` block if `contentId` is undefined.

also I put the "find a trigger" logic in a separate function.

### DIFF
--- a/test-support/helpers/ember-power-select.js
+++ b/test-support/helpers/ember-power-select.js
@@ -63,12 +63,16 @@ export function touchTrigger() {
 // Helpers for acceptance tests
 
 export default function() {
-  Test.registerAsyncHelper('selectChoose', function(app, cssPath, valueOrSelector) {
+  function findTrigger(cssPath) {
     let $trigger = find(`${cssPath} .ember-power-select-trigger`);
-
     if ($trigger === undefined || $trigger.length === 0) {
       $trigger = find(cssPath);
     }
+    return $trigger;
+  }
+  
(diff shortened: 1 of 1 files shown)


cibernox:
Can you explain a bit more how come the `aria-owns` of the trigger is not rendered yet? I think that it should.

noslouch:
given this template:
```handlebars
<div class="component-class">
{{#if promise.isFulfilled}}
  {{#power-select options=promise onchange=(action (mut activeThing)) as |thing|}}
    {{thing.name}}
  {{/power-select}}
{{else}}
  Loading...
{{/if}}
</div>
```
And then I'm running `selectChoose('.component-class', '.ember-power-select-option:eq(3)')`. 

Should I not use `.component-class` as the first argument?

cibernox:
I see. I think that this is expected behaviour. If instead of a select you had a link and you used `click('the link')`, this would also fail, am I right?

You cannot attempt to interact with an element that is not in the page yet, you should use `andThen` to wait for the promise to fulfill and then you can interact with the select.

cibernox:
I think that this is working as intended. Interacting with a select that is not yet in the screen is expected to fail.

PR #873, opened 2017-03-24
[selectSearch]: use element to open if not open close #868

### LINKED ISSUE #868: Test Helpers - pass a DOM element?
Now that selectSearch uses native-events, doing something like below does not work anymore, specifically b/c of this [line](https://github.com/cibernox/ember-power-select/blob/master/test-support/helpers/ember-power-select.js#L123).  What do you think about refactoring the selectSearch to also take a (potential) dom element?  I can submit a PR, but wanted to run it by you first before tackling.

```
  selectSearch('.t-address-country:eq(2)', keyword);
  selectChoose('.t-address-country:eq(2)', countryName);
```

The other option is to make sure it is open before doing `selectSearch` to bypass that if block, but then it seemed better to have selectSearch to also take a DOM element.

snewcomer:
I think this is the simplest solution.  Is it planned in the future that acceptance test helpers (find, click) will be jquery-free or can/will use both?  i.e. in the future, passing `:eq(2)` won't work?

close #868

### DIFF
--- a/test-support/helpers/ember-power-select.js
+++ b/test-support/helpers/ember-power-select.js
@@ -120,7 +120,7 @@ export default function() {
     let $content = find(`#${contentId}`);
     let dropdownIsClosed = $content.length === 0 || $content.hasClass('ember-basic-dropdown-content-placeholder');
     if (dropdownIsClosed) {
-      nativeMouseDown(triggerPath);
+      nativeMouseDown($trigger.get(0));
       wait();
     }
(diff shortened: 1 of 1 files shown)


snewcomer:
random fail?

cibernox:
I'm back from emberconf. I just found that I added a comment but it remained "pending" so I never sent it 🤦‍♂️ . 

Per my comment, i've created https://github.com/cibernox/ember-power-select/pull/882 to see if we can use the native DOM helpers before and then you could rebase your PR over that. Let's see how it goes.

PR #878, opened 2017-03-30
Allowing for Custom group component

alexander-alvarez:
This allows for functionality such as collapsible groups.

Let me know on guidance as to where to put things (docs vs cookbook, test placement)

@cibernox


addon/templates/components/power-select/options.hbs
```diff
@@ -5,18 +5,29 @@
...
-        options=(readonly opt.options)
-        select=(readonly select)
-        groupIndex=(concat groupIndex index ".")
-        optionsComponent=(readonly optionsComponent)
-        role="group"
-        class="ember-power-select-options" as |option|}}
-        {{yield option select}}
-      {{/component}}
-    </li>
+    {{#with
```
cibernox: I'm still jet-lagged AF, but wouldn't be easier to just render the default group component?

```hbs
{{#if (ember-power-select-is-group opt)}}
  {{#component groupComponent 
    group=opt 
    select=select
    optionsComponent=(readonly optionsComponent) as |option|}}
    {{yield option select}}
  {{/component}}
{{else}}
```

I see that with this approach the custom component only modifies the group, no the options. I like that tho. I'll review it again this week once I have a functional brain.

cibernox: After a second read, this starts to make sense tho.

alexander-alvarez: No problem...

I decided to keep them separate so you wouldn't have to worry about instantiating the options component in the group component when overriding the group component. All you have to worry about is adding a `yield` when overriding the group component, and this structure handles rendering the options component in place.

cibernox:
Reviewing this is on my schedule for this afternoon.

alexander-alvarez:
thanks.
looks like the tests fail inconsistently in master.

addon/templates/components/power-select/options.hbs
cibernox: What about this?

```hbs
  {{#component groupComponent group=opt extra=extra select=select as |option|}}
    {{#component optionsComponent
      options=(readonly opt.options)
      select=(readonly select)
      groupIndex=(concat groupIndex index ".")
      optionsComponent=(readonly optionsComponent)
      role="group"
      class="ember-power-select-options" as |option|}}
      {{yield option select}}
    {{/component}}
  {{/component}}
```

Seems simple enough and users that want to customize the group but not the options can still just `{{yield}}`. Is there any feature I'm missing?

alexander-alvarez: sure. It's just moving the args from hash to the component helper, but it's more legible, so I'll do that

cibernox: When it comes to documentation, it's tricky but what I think I'd do is:

1. As with all public API options, list it in the `API reference` section.
2. Given that this is a pretty advanced customization (and I imagine you are doing this because you need it), an entry in the cookbook with such example would be great.

tests/dummy/app/templates/components/collapsible-group.hbs
```diff
@@ -1,6 +1,6 @@
 <li class="ember-power-select-group" aria-disabled={{ember-power-select-true-string-if-present disabled}} role="option">
   <span class="ember-power-select-group-name collapsible {{if collapsed 'is-collapsed'}}" onclick={{toggle 'collapsed' this}}>{{groupName}}</span>
-  {{#unless collapsed}}
+  <div class="{{if collapsed 'hide-options'}}">
```
cibernox: I think that not rendering anything would be better for performance. You don't pay the price of rendering the options unless you uncollapse the group.

alexander-alvarez: I do too.
We were get exceptions though when using the up and down arrow keys, which now that I think about it highlights how the example isn't accessible. Investigating.

cibernox: I almost bet money that it's because of `https://github.com/cibernox/ember-power-select/blob/master/addon/components/power-select.js#L562`

alexander-alvarez: It is triggered by that. I'm investigating how we would enable being able to "select" a group when it's collapsed, so that you can trigger the expand via keyboard.
Would groups have to become an options when collapsed? You have any ideas about it?

cibernox:
Yes, I'd do that. Which is more or less the approach of http://www.ember-power-select.com/cookbook/navigable-select

alexander-alvarez:
so would we even need this `groupComponent`?

cibernox:
For having collapsable options probably no, but I think that it's still a good idea in case someone wants to mess with the markup of the groups

alexander-alvarez:
Got it. You want me to strip all the collapsible stuff from this PR, and propose an addition with collapsible groups to the Cookbook in another PR?

cibernox:
Yes, I think you can strip it. The docs should have an entire section dedicated about how to replace individual components with your own, but that's on me.

cibernox:
One last request, can you add a test that ensures that the `groupComponent` receives the `select`, `group` and `extra` properties? It's very easy to remove one of those by mistake.

alexander-alvarez:
@cibernox I added tests, but did it in a different manner -- let me know if how I did it is fine.

cibernox:
@alexander-alvarez That's actually clever. I like it. But you need to add the mandatory `onchange` function to those tests.

cibernox:
Approved, let's hope travis does it too

cibernox:
@alexander-alvarez There only one failing test that seems a real failure, other than that this looks perfect. Do you have time to fix it?

cibernox:
Thanks!

cibernox:
I'll try to get something else in today and release a new version.

alexander-alvarez:
yeah I don't know what was going on there... what i had before should have worked but ¯\_(ツ)_/¯

PR #889, opened 2017-04-14
Add `tagName` to arguments passed from `power-select-multiple` to `power-select`.

luketheobscure:
We reopen `power-select` so we can add some attribute bindings for test hooks. This requires all instances to not be tagless, so we need the `tagName` property we pass to `power-select-multiple` to trickle down to `power-select`.


luketheobscure:
In the end I fixed it with `PowerSelect.reopen({tagName: 'div')`, but it still seems like the missing `tagName` argument is an oversight.

cibernox:
LGTM. Thanks!

Issue #894, opened 2017-04-20
Dynamic sub menus - JS errors


Moxide:
Hello,
I'm trying to implement dynamic sub menus in a EPS : [See this twiddle](https://ember-twiddle.com/2f88479aa469d0a498acb4332fe13abb)
It's _mostly_ working, but JS errors are thrown when submenus are shown (`optionElement` being null).

Temporarily, I added a "not null" check on `optionElement` ([see this commit](https://github.com/Moxide/ember-power-select/commit/cbdd0086fdf61c0eaa7bce749b1cabe7344d1015)), but could anyone take a look ? I must have overlooked something, but cannot see where !

Thanks :-)

cibernox:
Very creative! Yes, I never accounted by the posibility that options could not be rendered. I'd take a PR doing that.

cibernox:
@Moxide There is a few details this could be polished. Arrow navigation mainly.

I think that the best UX if when an item of a subgroup gets highlighted, the group automatically opens. I'll draft something based on the twiddle you shared.

cibernox:
@Moxide I hacked this: https://ember-twiddle.com/9e0bb0fbff5283856661066d87deb043?openFiles=components.power-select-custom-group.js%2C

Moxide:
Great ! Thank you very much for the insights !
There are still some quirks. For instance, in my application, the scroll position is "jumping" when I mouse over a group that is close to the bottom of dropdown. I had to prevent the scrollbars to appear.

cibernox:
I saw something like that. I think that it's because the select tries assumes that elements are inside the list of options and if an option is below the limit of visible are, it scrolls the list to put it into the viewport. 

I'm not sure if this kind of component is better built on top of EPS or perhaps it's too different from a select and should be build on top of ember-basic-dropdown.

Either way, your commit adding an early exit if the item you want to scroll to is not present is good. PR it!

Moxide:
Done ;-)
Thanks for the explanation about autoscrolling.
In the application I'm working on, I make use of the search field in same time as the submenus, so in my case it's better to use EPS, isn't it ?

PR #895, opened 2017-04-21
Abort scrollTo if element to scroll is null

Moxide:
See #894


cibernox:
@Moxide LGTM. That failure seemed random. Triggered a rebuild.

cibernox:
Thanks!

PR #909, opened 2017-05-04
Add dev/test assertion to tell users that promises inside groups are not allowed.

cibernox:
/cc @offirgolan This would have helped you I think


offirgolan:
Thanks for putting this together 😄 . It will definitely help throughout the rest of my upgrade process!

PR #911, opened 2017-05-09
Fix momentum scrolling issue on iOS devices

### LINKED ISSUE #841: Scrolling on mobile
Scrolling on mobile is not as fluid as a native select or other packages. It is choppy and hard to scroll through results. Is this on the radar at all?

mike183:
This has been bugging me in my application for a while!

This resolves the momentum scrolling issues on iOS devices, have tested it on my iPhone 5 running iOS 10.3.1.

Fixes #841

### DIFF
--- a/vendor/ember-power-select-bootstrap.css
+++ b/vendor/ember-power-select-bootstrap.css
@@ -195,8 +195,8 @@
   user-select: none;
   -webkit-user-select: none; }
   .ember-power-select-options[role="listbox"] {
-    overflow-y: auto;
-    /* in firefox in windows this can cause a word-break issue. Try `overflow-y: scroll` if that happens */
+    overflow-y: auto; /* in firefox in windows this can cause a word-break issue. Try `overflow-y: scroll` if that happens */
+    -webkit-overflow-scrolling: touch;
     max-height: 14em; }
 
 .ember-power-select-option {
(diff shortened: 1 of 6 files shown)


cibernox:
@mike183 The only thing that concerns me it's the `overflow-y: scroll;` part. I believe that this will force the list of options to have a scrollbar on windows or osx (when using a mouse instead of a trackpad) even when the amount of options is small and there is no need to scroll.

Can you verify that? Would the `-webkit-overflow-scrolling: touch;` work with `scroll: auto`?

mike183:
> I believe that this will force the list of options to have a scrollbar on windows or osx (when using a mouse instead of a trackpad) even when the amount of options is small and there is no need to scroll.

This is correct

> Can you verify if I'm right? Would the -webkit-overflow-scrolling: touch; work with scroll: auto?

I'm pretty sure I've tried this in the past and it didn't work, though I will try again and report back.

vendor/ember-power-select-bootstrap.css
```diff
@@ -196,8 +196,13 @@
   -webkit-user-select: none; }
   .ember-power-select-options[role="listbox"] {
     overflow-y: auto;
-    /* in firefox in windows this can cause a word-break issue. Try `overflow-y: scroll` if that happens */
     max-height: 14em; }
+  @supports (-webkit-overflow-scrolling: touch) {
```
cibernox: Files in /vendor are autogenerated, but don't worry. They'll be overwritten the next time I publish a version.

mike183:
This is ready to go pending CI.

cibernox:
failure is unrelated.i'll check it

Issue #906, opened 2017-05-03
[RFC] Yielding components


NLincoln:
# Summary
Move to v2.0, allowing ember-power-select to shed it's ember 1.13 baggage and expose a nicer API

# Motivation
Your [emberconf keynote](https://www.youtube.com/watch?v=MpFudGJn2J0) got me thinking. `ember-power-select` was created back in ember 1.13.8, and despite dropping support for ember < 2.3.1 and gaining embers powerful component API's, ember-power-select has not been able to migrate to this new API, since it hit 1.0.0 while ember 1.13 support was still required.

However, as the addon has matured, it may be time to consider deprecating power-select 1.0 and 
moving towards a better API. This issue is to (hopefully) start that discussion and see if the community and users can come to a consensus. 

I think we all agree that while the power select API is very powerful, the popularity of this addon necessitates something far more flexible.

# Detailed Design
Most common usage:
```hbs
{{power-select
  options=options
  selected=selected
  onChange=(action (mut selected))
}}
```

This is roughly equivalent to the current invocation for `ember-power-select-blockless`, with the notable change of `onchange` becoming `onChange`, which makes it more in line with the other ember power components (inspiration from #899).

A `labelPath` can also be provided:
```hbs
{{power-select
  options=options
  selected=selected
  onChange=(action (mut selected))
  labelPath="foo"
}}
```

Again, nothing new. Where things get interesting is when we start passing blocks:

```hbs
{{#power-select ...args... as |select|}}
  {{#select.selected as |country|}} {{! names are subject to change }}
    {{country.name}}
  {{/select.selected}} 
  {{#select.option as |country|}}
    <img src={{country.flag}}>
  {{/select.option}}
{{/power-select
```

Other components could be yielded in a similar fashion, such as a `placeholder`.

## The path forward:
First, the `onchange` to `onChange` deprecation can be handled almost  immediately. 

After the community has reached consensus on the user facing API of the new power-select, we begin a beta series where we incrementally implement the new API.

# How we teach this
Obviously large amounts of changes to the docs will be required. Possibly provide v1 docs concurrently so users who have yet to migrate can switch over?

# Alternatives
* Wait for a way to incrementally deprecate current behavior to move into upstream ember, then wait for enough of the community to use it to warrant making it the minimum supported version.
* Of course, we could do nothing. (not desirable)

cibernox:
This is on the roadmap, indeed. I even bikeshedded a tentative API.

```
[...]

PR #923, opened 2017-05-28
WIP: Basic power select

NLincoln:
For now, this is very WIP, but it attempts to make a first stab at #906. 

- [x] existing tests pass
- [ ] Add new tests
- [ ] Docs


cibernox:
Hey, thanks for the effort. I need explore this a bit further later this week. It first glance something I'm not sure about is to expose the dropdown itself as a components, I think that exposing the power-select trigger should be enough.

```hbs
{{#power-select-basic as |select|}}
  {{#select.trigger as |selected|}}
    {{selected.name}} - <i class={{selected.icon-name}}></i>
  {{/select.trigger}}
  {{#select.dropdown}}
    {{#select.options as |opt|}}
      {{opt.name}}
    {{#select.options}}
  {{/select.dropdown}}
{{/power-select-basic}}
```

In the example above I try to not surface the distinction power-between select and basic-dropdown. 

To be completely honest, this is just one idea, probably working on it will make defects on it arise. I should be able to start doing more OSS soon and this is on the top of my list.

I need to catch a flight today, I'll be back on business in a couple days.

NLincoln:
Yeah I totally agree with you on the drop-down, don't worry it'll be fixed before all this is done. This initial work was more about making a proof of concept and getting the bulk of the code where I wanted it.

miguelcobain:
I'm excited with the customization possibilities this unlocks.

NLincoln:
Updated with your suggestions for the API. LMK what you think

cibernox:
I just fixed the tests. Can you rebase?

NLincoln:
Rebased 👍

miguelcobain:
What's the status of this? Any help needed?

NLincoln:
I've been pretty swamped lately with work stuff. Help would be greatly appreciated. 

TODO:

- [ ] Get the tests that are now failing passing again
- [ ] Docs
- [ ] Tests

Issue #919, opened 2017-05-26
Option select issue on iOS introduced in v1.7.0


jdurand:
*Issue introduced in **v1.7.0** and present in every subsequent versions*
*Reproduced on **iOS Safari***

The options wrapper `.ember-power-select-options` stays highlighted instead of immediately blurring out when an option is pressed. The option never selects unless it is carefully pressed for a few seconds and then released.

< 1.7.0:
![< 1.7.0](https://i.gyazo.com/a6279b450ab42307d283cd41ce70f403.gif)

\>= 1.7.0:
![>= 1.7.0](https://i.gyazo.com/8099071db7128f519f7513a2fe5f3bf8.gif)

Live example:
http://widgets.dev.libroreserve.com/booking/DEMO/DEMO

cibernox:
Sorry for the delay, I was on holidays.

On the link you posted I cannot se that select, but other selects seem to work just fine in chrome and safari, but I don't own an iOS device. However from your gif seems that you are using a mouse.

How can I reproduce that error?

I can't think in any change in 1.8 that would case such a big regression.

jdurand:
This works on most browsers and devices. The problem is only present on iOS devices.

I captured the gif Xcode's iPhone Simulator, that's why you see a mouse.

jdurand:
And the problem was introduced in 1.7.0.
I'll try dig a little more today.

cibernox:
Thanks. I'll also install xcode and the emulator. It's always good to have it.

MatthewPringle:
I can confirm this is a bug on the newer iOS. What is really strange is that some of the demo pages do not work yet this page does http://www.ember-power-select.com/docs/action-handling

MatthewPringle:
Ok think I might have figured this out. I can't test as I haven't ever used Power Select because of this bug.

So I wrote my own, and then had a bug, on device only, not simulator, where if focused on the input, clicking any of the list, with an action attached didn't seem to trigger the event. Seems like the same problem.

So anyway, my fix was to add an active class to the clickable options. That seems to get Safari's attention and everything works fine.

Can you try this @jdurand

cibernox:
> my fix was to add an active class to the clickable options

Can you expand on what you mean by that?

MatthewPringle:
OK, in my autocomplete, the code for the list / dropdown of results

```
{{#each results as |result|}}
        <li role="button" {{action "select" result.value result.label on="click"}}>{{result.label}}</li>
[...]

PR #931, opened 2017-06-08
Fix iOS issue with touchend event

jdurand:
Fixes https://github.com/cibernox/ember-power-select/issues/919

### DIFF
--- a/addon/components/power-select/options.js
+++ b/addon/components/power-select/options.js
@@ -74,7 +74,7 @@ export default Component.extend({
     this.element.addEventListener('touchend', (e) => {
       let optionItem = e.target.closest('[data-option-index]');
 
-      if (!optionItem || !(0 in optionItem)) {
+      if (!optionItem) {
         return;
       }
 
@@ -84,7 +84,7 @@ export default Component.extend({
         return;
       }
 
-      let optionIndex = optionItem[0].getAttribute('data-option-index');
+      let optionIndex = optionItem.getAttribute('data-option-index');
       this.get('select.actions.choose')(this._optionFromIndex(optionIndex), e);
     });
   },


cibernox:
@jdurand What I don't get is how this works in the other browsers. Chrome by example also has a native "closest" function.

luketheobscure:
Might be safer to program defensively since based on the discussion here it seems like `closest` will sometimes be returning an array, and sometimes an element. Maybe check if `optionItem` is an array before trying to access anything on it?

jdurand:
My assumption that `closest` was broken was wrong.
The issue is that `e.target.closest('[data-option-index]')` doesn't return an array; it returns an element. the `touchend` handler was expecting an array as is the case with `$(e.target).closest('[data-option-index]')`

cibernox:
According to https://developer.mozilla.org/en-US/docs/Web/API/Element/closest , `closest` always return an element, never an array. Which also makes sense to me, the `closest` element can only be one.

This change seems correct, what I'm not very sure is why chrome for android doesn't exercises this. I'll merge once CI pases.

jdurand:
@cibernox my guess is that this case is handled by the click handler on Android, otherwise it would fail the same way. Perhaps somebody with an Android phone could confirm this.

cibernox:
I have one. Can you point me to an failing example?

jdurand:
The language select here reproduces the issue on iOS: http://widgets.dev.libroreserve.com/booking/DEMO/DEMO/

cibernox:
@jdurand Minified it's almost impossible to replicate. But I'll merge it. It makes total sense.

cibernox:
@jdurand Released in 1.8.5.
@luketheobscure This also includes your latest PR 🙇

PR #935, opened 2017-06-15
Add isOpen to the API reference docs

maxkwallace:
To http://www.ember-power-select.com/docs/api-reference.

Not a lot of thought went into this, so feel free to close if this is the wrong thing to do or I'm wrong about something! Just noticed `isOpen` in my console, didn't see it in the docs, and figured it'd be fastest to open a PR.

Thanks so much for the addon :) its beauty brings me much joy.


cibernox:
Thanks!

PR #943, opened 2017-07-01
Fix Y axis validation on choose breaking tests. [#929]

### LINKED ISSUE #929: *Y axis* validation on `choose` could break acceptance tests.
The *Y axis* [validation](https://github.com/cibernox/ember-power-select/blob/master/addon/components/power-select.js#L276) on `choose` should be disabled for testing, or have the option to manually disable it. We have many acceptance tests that rely on `selectChoose`, a simple change on the layout design (a margin of `2rem` added to a parent where the EPS is) is now making all those test fail. Maybe there is no need of being that severe for the test env.

So, two suggestions (working on PR for this):
* Adding an option to toggle that validation.
* Making that 2px value customizable.

esbanarango:
This PR fixes #929. Now the Y axis validation on choose is disabled on testing.


cibernox:
@esbanarango Hey! I was on holidays when you opened the issue and I read it, but I completely forgot. I'll review it tomorrow.

Thanks for taking the time.

esbanarango:
Thank you @cibernox!!

cibernox:
@esbanarango Hey, I tried to create a reproduction but I couldn't? Can you show me how?

NLincoln:
I've also ran into this issue. Sometimes in integration tests the drop-down renders in really weird places, and sometimes with the first option of the power select over the trigger. 

I think that test still has that weird positioning. I will also try to get a good repro going.

cibernox:
@NLincoln I see how this can happen but I'm not sure if if the best way is to conditionally avoid some logic in test or modify `selectChoose` to not trigger this behaviour. 
That's why I'd like a reproduction first.

cibernox:
I'm in the middle of a moving so I'll have 0 time until next week. Please ping me again if I haven't responded by next Wednesday.

lukemelia:
I think this is affecting us as well.

esbanarango:
I've been trying to reproduce it again but seems like it has to be with some specific css values.

jamesarosen:
As noted on #929, my problem was that the trigger was scrolled out of view. I added

```js
$(trigger)[0].scrollIntoView();
```

to `selectChoose` and my tests started passing. We should be able to reproduce it with something like

```hbs
```

cibernox:
@jamesarosen your problem seems different and seems indeed a good thing to do. I'd take a separate PR doing that. It will probably be next to impossible to test tho, so don't bother too much on that

cibernox:
@jamesarosen just check if `scrollIntoView` is available in IE. If not, guard it.

cibernox:
@esbanarango A PR to scroll the trigger into view has been merged and released in 1.9.4. Can you check if this helps with your issue in any form. I'm not convinced it will, but it is worth trying.

esbanarango:
@cibernox Tests fixed and completely removed that special conditional!

cibernox:
@esbanarango I pulled your changes but I don't get exactly what that test is supposed to test. It passes with or without the code that verifies the Y axis.

PR #944, opened 2017-07-03
Use new shims

cibernox:
There might be a few left, but very few

### DIFF
--- a/addon/components/power-select-multiple.js
+++ b/addon/components/power-select-multiple.js
@@ -1,6 +1,6 @@
 import Ember from 'ember';
-import Component from 'ember-component';
-import computed from 'ember-computed';
+import Component from '@ember/component';
+import { computed } from '@ember/object';
 import layout from '../templates/components/power-select-multiple';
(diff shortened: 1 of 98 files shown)


mattmcmanus:
Hey @cibernox, sorry to bug you directly on a PR like this, but it's this PR that inspired the question. What's responsible for exposing the new shims? The version of ember in `ember-source`? I want to add snippets for the imports to https://github.com/mattmcmanus/atom-ember-snippets, but I want to make sure I document what's required for them to work. Thanks!

cibernox:
@mattmcmanus The new shims are part of ember-cli-babel >=6.6.0. They are not exactly shims as they are transformed in compile time to globals. 
Per example `import { alias } from '@ember/object/computed` is expanded to `Ember.computed.alias` in the places `alias` was used, and the import statement is removed from the code.

The rough plan is that once those imports are in `ember-source` for real, `ember-cli-babel` could detect the version of ember-source in use and don't do anything.

mattmcmanus:
Thanks for the explanation!

PR #945, opened 2017-07-06
also pass `index`

bekzod:


cibernox:
I'm not sure if this is a good idea, let me explain why.
This block is reused in two different places, the list of options and the trigger, and in each one of them there is two possible options.

### The list of options
In the list of options the index is evident that is it's position among the currently rendered results, right?
Well, not exactly, because when you are render selects with groups, this index will be the index within the group.

### The trigger
On single selects, this should probably be null. On multiple selects it could either be null/undefined or the index within the selections. 

I am not agains the idea, but I'm not sure what should we yield on each case.

bekzod:
@cibernox for my use-case I needed index, so submitted code seems to be working haven't tried with groups, but if you think this won't work out, free to close

cibernox:
Is not that it won't work, is that it won't work in every case. I'd like it to behave consistenly.

Consistently might mean two things:

- We pass the ["composed index"](https://github.com/cibernox/ember-power-select/blob/master/addon/templates/components/power-select/options.hbs#L30). Composed index is a dot-separated number (string) that represents the position within groups or sub-groups. By example, if an option is the third option of a group, and that group is the 5th option within the list, the composed index would be "5.3". If that groups was a subgroup of the group in the second option would be "2.5.3".

- The index of the option ignoring nesting. ASCII art ahead
```
├── Group 1
│   ├── option 1
│   ├── option 2
│   ├── Group 1.1
│   │   └── option 3
├── Option 4
├── Group 2
│   ├── option 5
│   ├── option 6
```
Essentially, which one should it be? And when this block is reused for within the trigger?

bekzod:
maybe then pass groupIndex and optionIndex separately, haven't tried groups so have shallow knowledge about it and its implementation

cibernox:
Both options have pros and cons. Sometimes the value you care about is the absolute index, ignoring groups. Sometimes, you want the group. Perhaps you might want both. If that is the case we could `{{yield option select (hash absolute=absoluteIndex nested=groupIndex)}}`

absolute index would be by example `6` (or `5` if zero-based) and group index would be `"3.2"` (or 2.1 if zero based). But this might come at some perf penalty.

cibernox:
I'm doing some triage of old issues and after some thought I believe I'm going to close this. I think it's better to not add a feature nobody requested yet. I also think that it could be done in user land with CSS very often or  with a helper like
```
{{#power-select options=options ... as |opt|}}
  {{index-of opt among=options}}. {{opt}}
{{/power-select}}
```

PR #949, opened 2017-07-10
Fix issue with this.element being null touchMoveHandler

jdurand:
Seen numerous times in production logs:
Chrome: `Cannot read property 'removeEventListener' of null`
Safari: `null is not an object (evaluating 'e.element.removeEventListener')`

Likely caused by some race condition when the dropdown is destroyed just before some callbacks are fired.


cibernox:
I'm in the middle of a moving so I'll have 0 time until next week. Please ping me again if I haven't responded by next Wednesday.

jdurand:
👍, FYI I'm running this in production from this same branch

jdurand:
ping!

cibernox:
Thanks for this also!

jdurand:
Hey, thanks for your work; we don't say that often enough!

Issue #806, opened 2017-02-01
[a11y] Refine ARIA implementation


steveszc:
The following items resulted from an accessibility audit performed by Interactive Accessibility on an app I'm working on. I'm happy to take on the bulk of the work outlined below, but some of these changes may require non-trivial changes to the architecture of some of the add-on's components. I imagine I may need some assistance to make some of these changes without introducing breaking changes to the API or visual styling.

<details>
<summary>What's broken? / why do these changes need to be made?</summary>
The main issue currently is that due to incorrect implementation of aria roles and attributes the power-select component is not very usable by a screen reader. For example as as screen reader user keys through the options, the currently focused option can not be announced, and the number of available options can not be announced. This should be considered a "critical" issue from an a11y perspective. The items outlined below *should* fix this. ARIA is a complicated beast so I'm open to feedback on the items below. I've tried to provide links to W3C WAI-ARIA docs justifying these changes. At the end of the day these changes will need to be tested across a range of screen readers to ensure it functions correctly.
</details>

## Work items

- [x] **basic-dropdown** Use aria-owns instead of aria-controls to associate the combobox with the listbox<details>
 https://www.w3.org/TR/wai-aria/roles#combobox
</details>

- [x] **basic-dropdown** aria-owns/aria-controls must reference an element that exists in the dom.<details>
Currently the trigger's `aria-controls` attribute references an ID of a dom node that does not exist when the power-select is in the collapsed state. This is considered invalid markup and can break screen readers. The controlled element (the listbox) should be maintained in the dom even when collapsed, it can be an empty `div`/`ul`. - https://www.w3.org/TR/wai-aria/states_and_properties#aria-owns
</details>

- [x] **basic-dropdown** aria-haspopup should not be used on the trigger<details>
`aria-haspopup` is an implicit property of a combobox. Explicitly adding aria-haspopup implicitly and incorrectly defines the element as an aria menu (similar in semantics to a right-click contextual menu or the menu ribbon in MS Word or an OS), which not the correct role for the trigger. This attribute is not necessary on a combobox. 
https://www.w3.org/TR/wai-aria/states_and_properties#aria-haspopup 
https://www.w3.org/TR/wai-aria/roles#combobox
</details>

- [x] **basic-dropdown** aria-pressed should not be used on the trigger.<details>
`aria-expanded` is sufficient to imply behavior to screen readers. `aria-pressed` is intended for toggle buttons such as checkboxes or tri-state buttons.
[...]

PR #953, opened 2017-07-21
A11y

camskene:
This PR addresses several of the issues outlined in #806. Work has been tested with VoiceOver and JAWS and is working pretty well in single and multiple modes. The groups implementations needs work, specifically the id passed to track the current item doesn't account for the decimal place. Perhaps we could pair on that one?

### DIFF
--- a/addon/components/power-select.js
+++ b/addon/components/power-select.js
@@ -211,6 +211,10 @@ export default Component.extend({
     return `${uniqueId}-${indexOfOption(results, highlighted)}`;
   }),
 
+  resultsCountId: computed('publicAPI.uniqueId', function() {
+    return `ember-power-select-results-count-${this.get('publicAPI.uniqueId')}`;
+  }),
+
   // Actions
   actions: {
     registerAPI(dropdown) {
--- a/addon/components/power-select.js
+++ b/addon/components/power-select.js
@@ -202,6 +202,10 @@ export default Component.extend({
       && (!this.get('search') || publicAPI.lastSearchedText.length > 0);
   }),
 
+  ariaAutocomplete: computed('searchEnabled', function() {
+    return this.get('searchEnabled') ? 'list' : 'none';
+  }),
+
   // Actions
   actions: {
     registerAPI(dropdown) {
--- a/addon/components/power-select.js
+++ b/addon/components/power-select.js
@@ -206,6 +206,11 @@ export default Component.extend({
     return this.get('searchEnabled') ? 'list' : 'none';
   }),
 
+  highlightedId: computed('publicAPI.highlighted', function() {
+    let { highlighted, results, uniqueId } = this.get('publicAPI');
+    return `${uniqueId}-${indexOfOption(results, highlighted)}`;
+  }),
+
   // Actions
   actions: {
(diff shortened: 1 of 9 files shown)


tests/integration/components/power-select/a11y-test.js
```diff
@@ -289,34 +289,6 @@ test('Multiple-select: The searchbox has type `search` and `aria-controls=<id-of
   assert.ok(/^ember-power-select-options-ember\d+$/.test(find('.ember-power-select-trigger-multiple-input').attributes['aria-controls'].value), 'The `aria-controls` points to the id of the listbox');
 });
 
-test('Single-select: The listbox has `aria-controls=<id-of-the-trigger>`', function(assert) {
```
camskene: These tests were removed as UL's no longer have the aria-controls attribute. 

Per #806 "Remove aria-controls from ul elements inside the listbox. This is invalid. These lists do not actually control the trigger."

addon/components/power-select/options.js
```diff
@@ -57,8 +58,8 @@ export default Component.extend({
   },
 
   // CPs
-  'aria-controls': computed('select.uniqueId', function() {
-    return `ember-power-select-trigger-${this.get('select.uniqueId')}`;
+  'aria-multiple': computed(function() {
```
snewcomer: Very very minor but I wonder if this could be `computed.bool('${isArray(this.get('select.selected'))}')`

addon/templates/components/power-select.hbs
```diff
@@ -13,6 +13,9 @@
   as |dropdown|}}
 
   {{#dropdown.trigger
+    role="combobox"
+    aria-autocomplete=(readonly ariaAutocomplete)
+    aria-activedescendant=(readonly highlightedId)
```
snewcomer: nice!

addon/components/power-select.js
```diff
@@ -188,6 +188,19 @@ export default Component.extend({
...
+
+  highlightedId: computed('publicAPI.highlighted', function() {
+    let { highlighted, results, uniqueId } = this.get('publicAPI');
+    return `${uniqueId}-${indexOfOption(results, highlighted)}`;
+  }),
+
+  resultsCountId: computed('publicAPI.uniqueId', function() {
+    return `ember-power-select-results-count-${this.get('publicAPI.uniqueId')}`;
+  }),
+
```
snewcomer: This PR is great!  My only real comment would be wrt to the computed cache breaking keys.  For example, I wonder if `searchEnabled` or `publicAPI.highlighted` or `publicAPI.uniqueId` was removed if your tests would still pass?

addon/components/power-select/options.js
camskene: That doesn't work.

cibernox:
Hey, this has been open for way too long. I'll try to allocate some time on Thursday for going though the changes and check how VoiceOver behaves.

rwwagner90:
@camskene I realize this is super old, but could you please rebase and fix conflicts?

mkszepp:
In PR #1690 we have improved / fixed some parts for A11y support. For more details see PR

PR #959, opened 2017-08-07
Ignore keydown with modifier keys pressed

espadrine:
If we do not ignore them, we end up, for instance, filtering by "r" when when we
press ctrl+r to reload the page, or filtering by "l" when doing ctrl+l to select
the browser's URL.

It is always a bit annoying to witness this involuntary filtering, but more so
when the widget auto-saves its content. Especially so with ctrl+r, with which
you might not even notice the change because of the page reload.

### DIFF
--- a/addon/components/power-select.js
+++ b/addon/components/power-select.js
@@ -291,6 +291,9 @@ export default Component.extend({
       if (onkeydown && onkeydown(this.get('publicAPI'), e) === false) {
         return false;
       }
+      if (e.ctrlKey || e.metaKey) {
+        return false;
+      }
       if (e.keyCode >= 48 && e.keyCode <= 90) { // Keys 0-9, a-z or SPACE
         this.get('triggerTypingTask').perform(e);
       } else if (e.keyCode === 32) {  // Space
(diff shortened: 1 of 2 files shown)


cibernox:
I'm a bit surprised, because on my pressing keys with modifiers do not generate any text on the focused input. The key combination is trapped by the browser/OS and it never updates the search box.

What browser/OS are you using to reproduce this?

espadrine:
@cibernox I'm not sure what you mean by "generating text".

The issue is that the selector filtered with "r" when typing ctrl+r. It doesn't input text.

For instance, on the test I added, we have a selector with "one", "two", "three", etc.
When we do ctrl+r (to, say, reload the page), without this patch, the selector selects "three" (the first option that contains an "r").

As for how to reproduce it in real life, I work on qonto.eu, an app which has a field that gets auto-saved. Additionally, it asks for confirmation when the user leaves or reloads the page.

I use Google Chrome, which does not stop the JS engine when the confirmation is shown.
As a consequence, the selector has time to perform filtering, select a different entry, and auto-save it (through an XHR), before the user clicks "yes, chrome, please do reload the page".

cibernox:
Oh, you are not talking about the searchbox but about the automatic selection when the component is closed (or highlight when it's open).

You're totally right, that is wrong. And your fix is the right one.

Thanks!

cibernox:
Released on 1.9.3! 🎉

espadrine:
Thanks a lot! Godspeed!

PR #960, opened 2017-08-09
selectChoose, selectSearch scrollIntoView

jamesarosen:
If the trigger is scrolled off screen, the options menu may not appear properly, causing the action not to fire. This commit changes the test helpers to scroll the trigger into view (assuming the browser supports the function) before trying to open the menu and select an item.

### DIFF
--- a/addon-test-support/helpers.js
+++ b/addon-test-support/helpers.js
@@ -82,6 +82,10 @@ export async function selectChoose(cssPathOrTrigger, valueOrSelector, optionInde
     }
   }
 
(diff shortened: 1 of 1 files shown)


jamesarosen:
`scrollIntoView` is [available](http://caniuse.com/#feat=scrollintoview) everywhere but in Opera Mini.

jamesarosen:
See also #929

cibernox:
Published in 1.9.4 🎉

PR #963, opened 2017-08-14
Feature - Add support for Numpad typing

cristi-badila:
This PR adds the ability for users to select an option by typing from their Numpad keys.

In order to achieve this I expanded the whitelist of allowed keycodes that trigger the `triggerTypingTask` task so as to include the appropriate keycodes. I also had to update the code which converted the keycode into an actual character so as to take this into account since the keycode for Numpad keys does not directly translate to the unicode representation of the equivalent number character. 

Docs:
- info on Numpad keys in Javascript [here](https://www.cambiaresearch.com/articles/15/javascript-char-codes-key-codes)
- info on String.fromCharCode [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/fromCharCode)

### DIFF
--- a/addon/components/power-select.js
+++ b/addon/components/power-select.js
@@ -294,7 +294,8 @@ export default Component.extend({
       if (e.ctrlKey || e.metaKey) {
         return false;
       }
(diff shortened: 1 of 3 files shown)


cibernox:
I'll release it today

PR #965, opened 2017-08-14
failing test

BenjaminHorn:


cibernox:
Included by #966 
Thanks for raising this. Even if it wasn't a real bug the added assertion will lead developers in the right direction.

PR #966, opened 2017-08-16
Raise helpful assertion

cibernox:
Replaces #965 

@BenjaminHorn It turns out there was no error. The problem is that when your options are not strings, Ember Power Select needs to know what field it should filter by, using `seachField="name"`. Since you didn't provide any, options were coherced to `[object Object]`, that is why `c` and `o` and `b` and `j` and `t` matched.

I've added a dev-time assertion that will be stripped in production builds to help users when they forget about `searchField`.


BenjaminHorn:
thanks @cibernox :+1:

PR #974, opened 2017-09-04
Allow null options

jakesjews:
Prevent assertions from being triggered by null options

### DIFF
--- a/addon/components/power-select.js
+++ b/addon/components/power-select.js
@@ -16,7 +16,8 @@ import {
   filterOptions,
   countOptions,
   defaultHighlighted,
-  advanceSelectableOption
+  advanceSelectableOption,
+  isGroup
 } from '../utils/group-utils';
 import { task, timeout } from 'ember-concurrency';
 
@@ -470,9 +471,8 @@ export default Component.extend({
       (function walk(collection) {
         for (let i = 0; i < get(collection, 'length'); i++) {
           let entry = collection.objectAt ? collection.objectAt(i) : collection[i];
-          let subOptions = get(entry, 'options');
-          let isGroup = !!get(entry, 'groupName') && !!subOptions;
-          if (isGroup) {
+          if (isGroup(entry)) {
+            let subOptions = get(entry, 'options');
             assert('ember-power-select doesn\'t support promises inside groups. Please, resolve those promises and turn them into arrays before passing them to ember-power-select', !subOptions.then);
             walk(subOptions);
           }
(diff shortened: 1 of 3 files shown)


cibernox:
I'm not sure this is a good idea. Why would you pass an array containing "holes"? That is almost certainly unintended. 
Also, with this solution, the component would render an option with it's own `<li>` that is empty, which I also think it's wrong. A null option should just be skipped.

I believe the right way of dealing with this to fail with a helpful error message so the developer makes sure of doing `options.compact()` to remove holes before passing them to the select.

If you want to improve or add assertions so it's clear for the developers that they should remove holes from the options array before passing it to the select, I'd take a PR for that.

cibernox:
Thanks for taking the time 🙇🏻
I just don't think handling null options this way is the right approach 🙁

jakesjews:
@cibernox we have several places in our app where we allow selecting nothing as an option which worked with previous versions. What should we do instead?

jakesjews:
I just realized my title was a bit misleading the intention was more so to allow a single null option not necessarily more than one

jakesjews:
@cibernox any chance you could take another look at this in light of the comments above?

jakesjews:
For the most part we don't have searches on the select with null options. An example of a typical dropdown we have is one which allows engineers to select the index of an array they want to use inside a programmable logic controller. So the options would end up being something like [null, 0, 1, 2, 3, 4, 5, 6, 7]. We could wrap all the selects but it seems a bit overkill when we just need a list of primitive unsearchable numbers. 

We also run into an issue where we want the select to represent a list of possible relationships for example temporarily assigning an assembly line spindle as a backup of another with the intention of removing the relationship later. In that case then we need to make wrappers around each ember data object like [{ value: null, name: 'nothing' }, { value: emberDataObject, name: emberDataObject.name }] which can also be problematic because we need to make sure websocket updates can also update the bound names accordingly so the wrappers would also need to have computed name properties.

I guess these things can be worked around but it seems like a lot of trouble to prevent a null check. This behavior did work in version 1.5.0 so maybe the null checks won't be very problematic?

cibernox:
> This behavior did work in version 1.5.0 
That is interesting, I'll investigate it. As far as I can remember, this was never expected to work.

jakesjews:
I mention it because we're stuck on 1.5.0 because of this

samselikoff:
Just chiming in that it would be nice to have a blessed path for this pattern - running into it myself. Use case is to let the user select "I don't know", but to have that map to a `null` value.

jakesjews:
@samselikoff I have a published fork eflex-ember-power-select which has this fix in it if you want to use it. I intend to keep it up to date with changes in this repository.

PR #976, opened 2017-09-08
Check for `isDestroying` instead of `isDestroyed`

luketheobscure:
I'm not able to reproduce outside of my application, but during teardown after an acceptance test `_updateOptionsAndResults` throws an error. At the time of the error, `isDestroyed` is false, but `isDestroying` is true. 

This seems like a safe change to make, but if you want I can update to: `if (get(this, 'isDestroyed') || get(this, 'isDestroying'))`.

### DIFF
--- a/addon/components/power-select.js
+++ b/addon/components/power-select.js
@@ -512,7 +512,7 @@ export default Component.extend({
   },
 
   _updateOptionsAndResults(opts) {
-    if (get(this, 'isDestroyed')) {
+    if (get(this, 'isDestroying')) {
       return;
     }
     let options = toPlainArray(opts);


addon/components/power-select.js
```diff
@@ -512,7 +512,7 @@ export default Component.extend({
   },
 
   _updateOptionsAndResults(opts) {
-    if (get(this, 'isDestroyed')) {
+    if (get(this, 'isDestroying')) {
```
jasonmit: `this.isDestroyed || this.isDestroying` should be sufficient and you should not need to go through Ember.get for either prop.

cibernox:
I neglected this for way too long, sorry.

WhiIe the fix seems reasonable, this symptom seems very weird to me. Are you doing something special? Are you returning promises from the `search` action? Using asynchronous ember-data relationships?

I'd like to understand what is going on. Checking for `isDestroyed` already felt dirty to me, and this feels even dirtier.

luketheobscure:
My turn to apologize for neglecting this. ;)

It seems I can no longer recreate the original problem. If I ever see it again I'll reopen with steps to reproduce.

PR #987, opened 2017-10-03
Don't validate Y axis in testing

cibernox:
Closes #943


cibernox:
@esbanarango Do you think this will fix the problem?
I was reluctant to disable the check in testing because I wanted to keep the test for the feature, but by making it a CP i can override it on that specific test.

PR #991, opened 2017-10-11
Allow negative dropdown-margin values

ktrhn:


cibernox:
That makes sense! Thanks!

PR #994, opened 2017-10-14
Delete residual CSS from tagName removal. Allow classNames propagation.

nathanhammond:
`classNames` being passed forward allows direct configuration of `{{basic-dropdown}}` which matters for `renderInPlace=true` as you may need to distinguish in CSS between an `.ember-power-select` and your own custom `.ember-basic-dropdown` on the same page.


cibernox:
Ignore the failure, I'll check it now.

Thanks!

PR #995, opened 2017-10-15
Expose placeholder to beforeOptionsComponent.

nathanhammond:
This allows behavior like having the placeholder serve as a header for the select when open.


cibernox:
Thank you sir 🎩

PR #996, opened 2017-10-17
Bump ember-cli, ember, ember-truth-helpers

mixonic:
Bump ember-cli, ember, and ember-truth-helpers.

The mostly cosmetic differences are from `ember init`, should be cleaner going forward.

ember-truth-helpers had a 2.0 release recently which bumped it to the new modules format. As this library has a dependency, we need to update here. The semver breaking change in the 2.0 release is that ember-truth-helpers no longer supports Node < 4 or Ember < 2.0. Changelog: https://github.com/jmurphyau/ember-truth-helpers/releases/tag/v2.0.0

ember-power-select already does not support Node < 4 b/c of its dependency on ember-cli-babel 6.6.0. If ember-power-select currently supports only Ember >= 2.0 then I do not suggest a major version bump. If it does support legacy Ember, this patch requires a major version bump dropping support.

Edit: I had to add a dev dependency on `testdouble`. Something in the `ember-cli-blueprint-test-helpers` seems to demand it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,6 @@
 
 # compiled output
 /dist
-/fastboot-dist
 /tmp
 
 # dependencies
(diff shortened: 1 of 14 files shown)


mixonic:
This passes! I would love to see it in a release before Ember 2.17 😁

cibernox:
In!

cibernox:
Thanks! I'll release in a moment. The minimum version of Ember is 2.3.1, so we're fine.

cibernox:
Released in `1.9.11`

PR #1000, opened 2017-10-21
Update tests

cibernox:


tests/integration/components/power-select/general-behaviour-test.js
```diff
@@ -891,23 +891,23 @@ module(
...
+      run(() => this.set('selected', promise1));
 
       await render(hbs`
         {{#power-select options=numbers selected=selected onchange=(action (mut foo)) as |option|}}
           {{option}}
         {{/power-select}}
       `);
 
-      this.set('selected', promise2);
+      run(() => this.set('selected', promise2));
```
rwjblue: hmm, this.set should have already been run wrapped, I don’t think you should have to do it again 🤔

cibernox: And it didn't make a difference. I was just trying things.

PR #1006, opened 2017-10-26
[FEATURE] Implement native-like behaviour when power-select is closed.

Serabe:
When a select element is closed and focus, repeating the first character over and over again, cicles through the results.

For example, given the list of options 1, 2, 20, 21, 22; pressing _2_ twice (in the conditions specified above) will select _20_ and not _22_.

While native behaviour is a bit more complex than currently implemented (for both closed and open states), this take us closer to the real behaviour while we look for an actual spec of it. The cycling of the matches array relates to this more complex behaviour, though for now we could just select the right option.

The function `onlyOneLetterRepeatedAtLeastOnce` function checks that the string is just that. Could not find a better name but I am very much open to suggestions. Using the regular expression is 20 - 30% slower by some quick perf tests but is left in the docs in case it helps to understand its behaviour (never thought I would say that).

### DIFF
--- a/addon/components/power-select.js
+++ b/addon/components/power-select.js
@@ -52,6 +52,36 @@ function toPlainArray(collection) {
   return collection.toArray ? collection.toArray() : collection;
 }
 
+/**
+ * Checks if a string is just one character repeated at least once.
+ *
+ * Same as /^(.)\1+$/ but faster.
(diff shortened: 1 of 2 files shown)


Serabe:
New PR coming soon.

PR #1013, opened 2017-11-09
hotfix for error calling set on destroyed object during tests

joeypedicini92:
during our test runs, we would randomly get the "error calling set on destroyed object" message and there was no real way to track it down since the error was coming from the updateState function on the power select component itself. 

This fix just checks that the component is not destroyed before trying to call set

### DIFF
--- a/addon/components/power-select.js
+++ b/addon/components/power-select.js
@@ -650,11 +650,15 @@ export default Component.extend({
   },
 
   updateState(changes) {
-    let newState = set(this, 'publicAPI', assign({}, this.get('publicAPI'), changes));
-    let registerAPI = this.get('registerAPI');
-    if (registerAPI) {
-      registerAPI(newState);
+    if(this.isDestroyed) {
+      console.log('tried to set a destroyed ember power select object')
+    } else {
+      let newState = set(this, 'publicAPI', assign({}, this.get('publicAPI'), changes));
+      let registerAPI = this.get('registerAPI');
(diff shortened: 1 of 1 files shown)


cibernox:
This is weird. This fix should not be needed. Before accepting this, could you give me a stack trace of from where is this being called? I suspect the check should be upper in the call stack.

joeypedicini92:
I reverted back to 1.9.6 which is what we were on previously and now I can't reproduce the error. I did notice in my package-lock file that the ember-basic-dropdown changed from 0.33.5 to 0.33.9, that's the only other package change. I can try going back down to 0.33.5 and see if the error pops back up and let you know

joeypedicini92:
I figured it out...If I interact with the browser at all while the tests are running the error comes up. Has nothing to do with versions. However, with this fix, even when I interact with the browser the error doesn't pop up.

cibernox:
I see. However, I'd like to better understand the cause. I sounds to me like an event not being properly torn down when the selects are destroyed.
The reason why I don't want to bindly take this fix is because I believe it's a symptom of a deeper problem somewhere else, as this should never be called on a destroyed select.

cibernox:
@joeypedicini92 I suspect your problem was fixed in #1023.

Can you verify? I'm going to close this but please reopen if the bug persists.

joeypedicini92:
@cibernox great thanks! i'll check it out when i get some time

PR #1023, opened 2017-11-15
Bail out of _updateOptionsAndResults if `isDestroying`

luketheobscure:
I opened a similar PR while back, but declined when I couldn't consistently reproduce the issue. Now I can, although the setup is hard to explain. The specific error I get is `TypeError: Cannot read property 'isOpen' of undefined`, which happens on line 540 (`publicAPI.isOpen`).

I have a page with two ember-power-selects. When the first is selected, it populates the choices of the second (from a `hasMany` association from the selected model). When ember tears down all the models at the end of an acceptance test, it triggers the change event on the `hasMany` relationship, causing ember-power-select to want to update, but since everything is mid-teardown I get the error I mentioned. There's a lot of other stuff going on in my test, but I think it boils down to a race condition that this PR fixes.

### DIFF
--- a/addon/components/power-select.js
+++ b/addon/components/power-select.js
@@ -525,7 +525,7 @@ export default Component.extend({
   },
 
   _updateOptionsAndResults(opts) {
-    if (get(this, 'isDestroyed')) {
+    if (get(this, 'isDestroying')) {
       return;
     }
     let options = toPlainArray(opts);


cibernox:
Do you know if `isDestroying` becomes `false` once the destroy process is completed? If so, it could cause problems. I'll check this further today 👍

luketheobscure:
The docs don't say definitively, but I did some spot checks and all the objects where `isDestroyed` was `true` also has `isDestroying` set to `true`.

I've been trying to get a simple example together, and it's been maddeningly difficult. I did discover that it requires the inverse relationship to exist on the `hasMany` models. The error occurs when the child models teardown, which update the parent model, which updates the ember-power-select that's trying to display the child models. 😓

cibernox:
I've investigated and you fix is correct. `isDestroying` is a flag that is set at the beginning of the runloop in which the component is scheduled to be destroyed, and `isDestroyed` is added at the end of that operation. There is no point in update anything when the component is about to die.

Thanks!

PR #1029, opened 2017-11-23
Enable native-like search logic

Serabe:
**NOTE** For this behaviour to work when the select is open, the user
needs to disable the search.

About this PR
======

This PR aims to give EPS a native-like behaviour.

What's the UX in current browsers?
======

For this, we compared safari, chrome, and firefox. Both safari and
chrome have different behaviours depending on whether the select is
open or closed. Firefox does not. We believe the select should behave
the same in both states, and a deeper look at blink's code seem to proof
it. For the purposes of this PR, the behaviour explained and implemented
is the one reproducible when the select is closed in safari and chrome
(firefox only implement such behaviour).

When a user focuses on a select, does not matter the state, they can
go to the next option that matches the treated input. The value the user
enters does not need to be the one that is used for matching, but
whatever the user types, it might not be what it's used for looking for
matches. The system keeps its 1s memory before forgetting the previous
search term (as it does now). Furthermore, the threshold in EPS is the
same we can find in webkit and blink.

The differences that this PR implements are two: search offsets and
cycling through the results. These behaviour are explained in depth
next.

Search offset
======

The original implementation applies an offset when looking for matching
results. If we are in _cycling mode_ (see next part) or starting a new
search (by focusing or not typing for 1s), the select will start looking
one result after the current highlighted / selected option. That means
that if we have selected an option that matches our current intention,
the user will not stay there, but go to the next option (if any). Once
you think about it, seems natural that if the user is interacting with
the component, it is because it is not interested in the current
selection.

On the other hand, if the user is typing a longer search term, the
offset can be 0, since the current option might be a valid one (changing
the original selection was already done when typing the first character,
see paragraph above).

In any case, current highlighted / selection index is added to the offset
above.

Cycling through the results
======

When a user focuses on a select and types a letter, let's say _a_, the
select will show the first result that matches _a_. If the user, within
the time threshold, types _a_ again, the select will go to the next
result that matches _a_ and not _aa_. That way, typing again the same
first letter several times let us cycle through the results.

It is important to note that this only works for the first character.
For example, _Unittt_ will not cycle through the result for _Unit_ in
a country selection.

Also, the repetitions need to be remembered. In a select full of names,
typing _aar_ will take us to the right selection (_Aaron_) and not to
a result for _ar_.

Show me the (original) code!
======

> Files linked here are pinned to the latest commit they were changed in.

The inital research was done via manual testing. Then, after inspecting
[blink's HTMLSelectElement
implementation](https://github.com/ChromiumWebApps/blink/blob/f26b5a9078d12d954a236c557776f4b1395726c6/Source/core/html/HTMLSelectElement.cpp), I studied
[blink's
TypeAhead](https://github.com/ChromiumWebApps/blink/blob/b6720b1767fb5208aa421237c087f062093929e7/Source/core/html/forms/TypeAhead.cpp). Current implementation is thought as a mix of
[TypeAhead::handleEvent](https://github.com/ChromiumWebApps/blink/blob/b6720b1767fb5208aa421237c087f062093929e7/Source/core/html/forms/TypeAhead.cpp#L60)
and
[filterOptions](https://github.com/cibernox/ember-power-select/blob/6b1e472d394a9322f9a25a668cf0686bbeb2ad2c/addon/utils/group-utils.js#L74-L95).

I stated above that the lack of cycling behaviour when the select is
open is quite likely a bug because of
[this comment that can be found in blink's
PopupListBox](https://github.com/ChromiumWebApps/blink/blob/f26b5a9078d12d954a236c557776f4b1395726c6/Source/web/PopupListBox.cpp#L320-L321).

Native matcher
======

The native matcher behaviour is just to look for a prefix.

```js
import { stripDiacritics } from 'ember-power-select/utils/group-utils';

function nativeMatcher(value, text) {
  return stripDiacritics(value).toUpperCase().startsWith(stripDiacritics(text).toUpperCase()) ? 1 : -1;
}
```

A new option has been added to tweak this behaviour: `typeAheadMatcher`.

Live example
======

You can find the behaviour of this PR in http://power-test.surge.sh/

- Try looking for _Aaron_
- Cycling behaviour is better done with the "F" (shorter and the group
in the second select is repeated).

TODO
======

- [x] Tests for power-select
- [x] Tests for filterWithOffset
- [x] Refactor filterWithOffset to findOption?

### DIFF
--- a/addon/components/power-select.js
+++ b/addon/components/power-select.js
@@ -11,6 +11,7 @@ import { isArray as isEmberArray } from '@ember/array';
 import ArrayProxy from '@ember/array/proxy';
 import layout from '../templates/components/power-select';
 import fallbackIfUndefined from '../utils/computed-fallback-if-undefined';
+import optionsMatcher from '../utils/computed-options-matcher';
 import {
   defaultMatcher,
   indexOfOption,
@@ -164,17 +165,9 @@ export default Component.extend({
     }
   }),
 
-  optionMatcher: computed('searchField', 'matcher', function() {
-    let { matcher, searchField } = this.getProperties('matcher', 'searchField');
-    if (searchField && matcher === defaultMatcher) {
-      return (option, text) => matcher(get(option, searchField), text);
-    } else {
-      return (option, text) => {
-        assert('{{power-select}} If you want the default filtering to work on options that are not plain strings, you need to provide `searchField`', matcher !== defaultMatcher || typeof option === 'string');
-        return matcher(option, text);
-      };
-    }
-  }),
+  optionMatcher: optionsMatcher('matcher', defaultMatcher),
+
+  typeAheadOptionMatcher: optionsMatcher('typeAheadMatcher', defaultTypeAheadMatcher),
 
   concatenatedTriggerClasses: computed('triggerClass', 'publicAPI.isActive', function() {
     let classes = ['ember-power-select-trigger'];
@@ -511,7 +504,7 @@ export default Component.extend({
   },
 
   filterWithOffset(options, term, offset, skipDisabled = false) {
-    return filterOptionsWithOffset(options || [], term, this.get('typeAheadMatcher'), offset, skipDisabled);
+    return filterOptionsWithOffset(options || [], term, this.get('typeAheadOptionMatcher'), offset, skipDisabled);
   },
 
   updateOptions(options) {
--- a/addon/components/power-select.js
+++ b/addon/components/power-select.js
@@ -409,6 +409,7 @@ export default Component.extend({
     // In general, a user doing this interaction means to have a different result.
     let searchStartOffset = 1;
     let publicAPI = this.get('publicAPI');
+    let repeatingChar = publicAPI._repeatingChar;
     let charCode = e.keyCode;
     if (this._isNumpadKeyEvent(e)) {
       charCode -= 48; // Adjust char code offset for Numpad key codes. Check here for numapd key code behavior: https://goo.gl/Qwc9u4
@@ -427,9 +428,9 @@ export default Component.extend({
       // If the term is longer than one char, the user is in the middle of a non-cycling interaction
       // so the offset is just zero (the current selection is a valid match).
       searchStartOffset = 0;
-      this.updateState({ _repeatingChar: '' });
+      repeatingChar = '';
     } else {
-      this.updateState({ _repeatingChar: c });
+      repeatingChar = c;
     }
 
     // When the select is open, the "selection" is just highlighted.
@@ -443,7 +444,7 @@ export default Component.extend({
 
     // The char is always appended. That way, searching for words like "Aaron" will work even
     // if "Aa" would cycle through the results.
-    this.updateState({ _expirableSearchText: publicAPI._expirableSearchText + c });
+    this.updateState({ _expirableSearchText: publicAPI._expirableSearchText + c, _repeatingChar: repeatingChar });
     let matches = this.filterWithOffset(publicAPI.options, term, searchStartOffset, true);
     if (get(matches, 'length') > 0) {
       let firstMatch = optionAtIndex(matches, 0);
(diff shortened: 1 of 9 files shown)


addon/components/power-select.js
```diff
@@ -402,14 +404,43 @@ export default Component.extend({
...
+    }
+
+    // When the select is open, the "selection" is just highlighted.
+    if (publicAPI.isOpen && publicAPI.highlighted) {
+      searchStartOffset += indexOfOption(publicAPI.options, publicAPI.highlighted);
+    } else if (!publicAPI.isOpen && publicAPI.selected) {
+      searchStartOffset += indexOfOption(publicAPI.options, publicAPI.selected);
+    }
+
+    // The char is always appended. That way, searching for words like "Aaron" will work even
```
cibernox: The only suggestion on the code is trying to batch changes and make a single call to `this.updateState` if possible. In this case it seems that you could do it.

addon/utils/group-utils.js
```diff
@@ -71,6 +71,50 @@ export function optionAtIndex(originalCollection, index) {
...
 
+function copyGroup(group, suboptions) {
+  let groupCopy = { groupName: group.groupName, options: suboptions };
+  if (group.hasOwnProperty('disabled')) {
+    groupCopy.disabled = group.disabled;
+  }
+  return groupCopy;
+}
+
+export function filterOptionsWithOffset(options, text, matcher, offset, skipDisabled = false) {
```
cibernox: Perhaps this function could be unified with the regular filter? Not needed, just wondering.

tests/integration/components/power-select/type-ahead-test.js
```diff
@@ -0,0 +1,170 @@
...
+    assert.ok(find('.ember-power-select-dropdown'), 'The dropdown is still open');
+  },
+
+  valueAt(coll, _idx, isGrouped = false) {
+    let idx = isGrouped ? _idx : _idx % coll.length;
+    return optionAtIndex(coll, idx).option;
+  }
+};
+
+function type(trigger, str, times = 1) {
```
cibernox: nit pick: Rename to `typeText` or `typeString`. Otherwise it seems a noun to me, synonym of "kind"

tests/integration/components/power-select/type-ahead-test.js
```diff
@@ -0,0 +1,170 @@
...
+};
+
+const WITH_EPS_OPEN = {
+  beforeInteraction(trigger, assert) {
+    clickTrigger();
+    assert.ok(find('.ember-power-select-dropdown'), 'The dropdown is open');
+  },
+
+  checkSelectedValue(value, trigger, assert) {
+    assert.equal(find('.ember-power-select-option[aria-current=true]').textContent.trim(), value);
```
cibernox: Could you also assert that the content of the trigger hasn't changed? It's important that when it's open it highlights but doesn't select.

Serabe: Done.

tests/integration/components/power-select/type-ahead-test.js
Serabe: Done.

addon/components/power-select.js
```diff
@@ -402,14 +406,45 @@ export default Component.extend({
...
+      term = c;
+    } else {
+      term = publicAPI._expirableSearchText + c;
+    }
+
+    if (term.length > 1) {
+      // If the term is longer than one char, the user is in the middle of a non-cycling interaction
+      // so the offset is just zero (the current selection is a valid match).
+      searchStartOffset = 0;
+      this.updateState({ _repeatingChar: '' });
```
cibernox: Can you coalesce this `this.updateState` with the one in line 446?

Serabe: Done

addon/utils/group-utils.js
Serabe: I removed the function and added `findOptionWithOffset`.

Serabe:
http://power-test.surge.sh/ is updated as well.

cibernox:
Thanks! This kicks in 2.0

PR #1039, opened 2017-12-08
updating ember-native-dom-helpers - fixes #1038

### LINKED ISSUE #1038: i don't know why all my tests that uses selectChoose are broken now ...
8 days ago i have updated my application and deployed it with ember-power-select 1.10.4 (with the test-support depreciation) All of my tests worked great !)
But now, and i really don't know why, when i test my application, all selectChoose calls fail ...

I have made a lot of tests during this day : 

change node version 6.10, 6.11, 6.10, 8 ... no effect
relaunch the last succed Travis build (fails now .. grr)
change/force  a specific ember-power-select version ... no effect

i have debugged the selectChoose function, and my analyse is that the dropdown content is not generated, why ? i don't know ...

``let options = findAll(`#${contentId} .ember-power-select-option`);``
`options` is empty, the previous click call (force dropdown opening) has no effect ..

My application works perfectly on Chrome,Firefox, Safari ... only the tests fail ...

Tomorrow i will test with an empty new application ..

Encountered errors come from the error throws from `selectChoose` function at line 128 and from the function `querySelectorAll` due to the call launchs from the line 114 ``potentialTargets = findAll(`#${contentId} ${valueOrSelector}`);`` because `valueOrSelector` is just an option value in my tests

mansona:


mansona:
unfortunately because this is actually a devDependency it's not going fix anything for consuming apps, it only worked for us in npm link mode because of the different semantics of npm in that context 😞

cibernox:
I've reverted the change in ember-native-dom-helpers that caused problems. I'll ensure EPS is prepared for the change before revert the revert.

PR #1042, opened 2017-12-15
Added new option allowEmptySearch

shijistar:
Added a new option `allowEmptySearch`, to ensure `search` method is still be called even when the search box is empty.

The background case is like that. We are having a large array to list in power-select. If we show all of them when first open the panel, the performance is bad. We decide to use a custom `search`  to only show top 20 matched items, even when the search box is empty. But right now, when search box is empty, power select automatically skip calling search. That's why we need this option.

### DIFF
--- a/addon/components/power-select.js
+++ b/addon/components/power-select.js
@@ -79,6 +79,7 @@ export default Component.extend({
   matchTriggerWidth: fallbackIfUndefined(true),
   preventScroll: fallbackIfUndefined(false),
   matcher: fallbackIfUndefined(defaultMatcher),
+  allowEmptySearch: fallbackIfUndefined(false),
   loadingMessage: fallbackIfUndefined('Loading options...'),
   noMatchesMessage: fallbackIfUndefined('No results found'),
   searchMessage: fallbackIfUndefined('Type to search'),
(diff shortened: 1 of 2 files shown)


cibernox:
If that is what you want, I'd implement it like this:

```hbs
{{#power-select options=first20opts selected=selected search=(action "search) do |opt|}}
  {{opt}}
{{/power-select}}
```

shijistar:
Well, thanks for you quick reply. Actually we also have another case, we have to do some complex query in server side while user inputting. It's something similar to `auto-complete`. We also use a custom `search` method, and we still want to show something even when user input nothing. But again if search box is empty, it doesn't perform search. Do you have good way for that. Thank you again.

cibernox:
The code pasted above already does that:
1) If the user didn't type anything, the initial `options` will be rendered
2) If the user types something, a search is performed and the result of that search is displayed
3) If the user deletes the search, the initial options will be displayed again, like if the user never typed.

If you don't want to perform a query to load the initial options until the user opens que select for the first time, you can lazily initialize it using `onopen`:

```hbs
{{#power-select options=initialOptions selected=selected search=(action "search) onopen=(action "loadInitialOptions") do |opt|}}
  {{opt}}
{{/power-select}}
```

```js
actions: {
  loadInitialOptions() {
    this.set('initialOptions', this.queryServerForDefaultOptions());
  }
}
```

PR #1046, opened 2018-01-05
fix: typo in "The Search" section

knownasilya:


mkszepp:
Unfortunately I can no longer merge this fixed with #1635.
Thasnk for reporting

PR #1049, opened 2018-01-16
Allows to customise the caret of the trigger

xcambar:
This PR is on the same track as the customisation od the Placeholder component, but dedicated to the caret (right-hand side arrow).

Our use case for this feature is the ability to use an `x` sign instead to clear the selection, which implies:
* changing the design (hence markup etc.)
* binding actions

### DIFF
--- a/addon/components/power-select.js
+++ b/addon/components/power-select.js
@@ -94,6 +94,7 @@ export default Component.extend({
   triggerComponent: fallbackIfUndefined('power-select/trigger'),
   searchMessageComponent: fallbackIfUndefined('power-select/search-message'),
   placeholderComponent: fallbackIfUndefined('power-select/placeholder'),
+  caretComponent: fallbackIfUndefined('power-select/caret'),
   buildSelection: fallbackIfUndefined(function buildSelection(option) {
     return option;
   }),
--- a/addon/components/power-select/caret.js
+++ b/addon/components/power-select/caret.js
@@ -1,0 +1,7 @@
+import Component from '@ember/component';
+import layout from '../../templates/components/power-select/caret';
+
+export default Component.extend({
+  layout,
+  classNames: ['ember-power-select-caret']
+});
--- a/addon/templates/components/power-select-multiple.hbs
+++ b/addon/templates/components/power-select-multiple.hbs
@@ -35,6 +35,7 @@
     optionsComponent=optionsComponent
     placeholder=placeholder
     placeholderComponent=placeholderComponent
+    caretComponent=caretComponent
     preventScroll=preventScroll
     registerAPI=(readonly registerAPI)
     renderInPlace=renderInPlace
@@ -95,6 +96,7 @@
     optionsComponent=optionsComponent
     placeholder=placeholder
     placeholderComponent=placeholderComponent
+    caretComponent=caretComponent
     preventScroll=preventScroll
     registerAPI=(readonly registerAPI)
     renderInPlace=renderInPlace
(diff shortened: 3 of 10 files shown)


cibernox:
I'm sitting on the fence about whether or not this is a request common enough to justify another extra component for the 100% of the userbase.

Do you think you could accomplish the same thing by overriding the entire trigger even if it's a bit more involved. 

I'm also curious what you want to do with the caret that you can't do with regular CSS.

xcambar:
At the moment, we copy/paste the entire standard trigger component into our codebase just to change the last line of the template (or, the carpet).

We do it because our design guidelines and UI framework require that a select is "clearable" with a right-hand side cross, which happens to be a SVG icon, not CSS.

Would you prefer a snippet?

cibernox:
For clearing the selected value you have the `allowClear=true`option. See the last section in http://ember-power-select.com/docs/the-trigger

Is that different from what you want?

xcambar:
Still yes, at least for we need to apply our own design to it, add accessibility roles etc.

The discussion here is about the caret but the deeper consideration (IMHO) is "Should EPS be more flexible in terms of design? And if so, how?"

I am submitting only the caret here because it is the smallest, easiest thing to kickstart the discussion. But in our journey to having EPS work within our design system, there's a lot I'd like to contribute back beyond the caret, if that's a direction you'd like the project to take. :D

cibernox:
For the design, I previously used inline svg inside the CSS to customize the look of the icon. Easy enough. IDK if that would suffice. 

However if you want a button to clear a selection, It seems to me that what you should be customizing is not the caret but the `".ember-power-select-clear-btn"`. If you have any problem with the aria roles I'd like to hear about it because it should be fixed by default.

xcambar:
As soon as I can, I'll share with you an implementation of a custom-designed EPS, it will be more explicit how some parts of EPS are unavailable for customisation.

Meanwhile, feel free to close this PR.

cibernox:
Ok. I'm travelling this week so it might take longer for me to get back to you.

patrickberkeley:
FWIW being able to customize the trigger caret would be useful for us as well. We use svg icons, but store them with https://github.com/BrianSipple/ember-cli-svgstore, so background image data uri in CSS would create duplication.

simonihmig:
I was thinking about providing basically the same PR here, luckily I found this old one first...

> FWIW being able to customize the trigger caret would be useful for us as well. We use svg icons, but store them with https://github.com/BrianSipple/ember-cli-svgstore, so background image data uri in CSS would create duplication.

Same here (using `ember-svg-jar`). Also I tried using it as a background data uri, but this does not work in my case, as I needed to use a dynamic color (CSS custom property or something like `currentColor`), but this doesn't work as a background image. 

I feel quite bad about *extending* the trigger component as [suggested here](https://github.com/cibernox/ember-power-select/issues/1174#issuecomment-438206348). I think this can get very fragile, as you are basically making the whole component implementation a public API. I think the only public API of a component should be how you invoke it in a template. For example you wouldn't be able to refactor an Ember component to a Glimmer component, when you expect users to have extended it. For this very reason we issue a warning when detecting users to extend from any ember-bootstrap component. This allowed us to refactor component after component to Glimmer components, without any breaking changes. 

In this case the `this.clear` action [used here](https://github.com/cibernox/ember-power-select/blob/master/addon/components/power-select/trigger.hbs#L8) would implicitly become a public API, as any extending component would rely on it, so you cannot rename or remove it without breaking those.

Sorry if this got a bit off-topic, but it explains why I would rather pass my own component than to extend an existing one.

patrickberkeley:
> this can get very fragile, as you are basically making the whole component implementation a public API

We have experienced the fragility you mention, particularly when we do upgrades. We have extended a number of `ember-power-*` public files because it was either that or maintain our own fork. The former seemed like the simpler maintenance route.

PR #1060, opened 2018-02-08
selectedItemComponent into beforeOptionsComponent

nathanhammond:
Make `selectedItemComponent` available in `beforeOptionsComponent`. This will allow for a persistent header showing the overall selected items inside of `power-select-multiple`. This matters in the case where the list ends up scrolling.


cibernox:
The bug was in master. Fixed.

cibernox:
Published as part 2.0.0-beta.2

PR #1086, opened 2018-04-08
Import click and fillIn from ember-native-dom-helpers / Fix #1085

### LINKED ISSUE #1085: Version > 2.0.0 / Must setup rendering context
Version > `2.0.0`

```js
import { selectChoose } from 'ember-power-select/test-support';
// ...
await selectSearch('[data-test-custom-field="Host"]', 'Carlos Doe');
// ...
```
image 2018-04-08 at 12 04 43 am

esbanarango:
Importing `click` and `fillIn` directly from `@ember/test-helpers` causes #1085. This change was introduced on Version `2.0.0-beta.3`. Here: https://github.com/cibernox/ember-power-select/commit/87b898f40bf5909d4149d7891f084bcbff352ec1#diff-aeaa9ea987e086a3faddb603853d857aR1

### DIFF
--- a/addon-test-support/index.js
+++ b/addon-test-support/index.js
@@ -1,5 +1,5 @@
-import { click, fillIn, settled } from '@ember/test-helpers';
-import { find, findAll } from 'ember-native-dom-helpers';
+import { settled } from '@ember/test-helpers';
+import { click, fillIn, find, findAll } from 'ember-native-dom-helpers';
 import { warn } from '@ember/debug';
 
 export async function selectChoose(cssPathOrTrigger, valueOrSelector, optionIndex) {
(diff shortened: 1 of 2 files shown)


cibernox:
I intentionally wanted to remove `ember-native-dom-helpers` and use `@ember/test-helpers`. The only apps have to do is be on a "recent" version of `ember-cli-qunit` or `ember-cli-mocha`, even if the apps don't want to use any of the new features in them.

cibernox:
I wonder if there is a way to make that explicit in the package.json

esbanarango:
@cibernox that's weird, we're actually on `"ember-cli-qunit": "^4.3.2",` but we still getting the `Must setup rendering context before attempting to interact with elements` error, and the only way we have solved it is by using `ember-native-dom-helpers`.

PR #1097, opened 2018-04-19
fixes nested native getter

### LINKED ISSUE #1094: Error with ember 3.1
Hi, I recently updated ember to version 3.1 and now ember-power-select is not working if I use the selected option.

```htmlbars
{{#power-select options=layouts
    selected=selectedLayout
    onchange=(action (mut selectedLayout)) as |layout|}}
  {{layout.name}}
{{/power-select}}
```

layouts is the result of
```javascript
this.get('store').findAll('layout')
```
The error message is 

```
Error: Assertion Failed: You attempted to access the `then` property (of <(unknown mixin):ember757>).
Since Ember 3.1, this is usually fine as you no longer need to use `.get()`
to access computed properties. However, in this case, the object in question
is a special kind of Ember object (a proxy). Therefore, it is still necessary
to use `.get('then')` in this case.

If you encountered this error because of third-party code that you don't control,
there is more information at https://github.com/emberjs/ember.js/issues/16148, and
you can help us improve this error message by telling us more about what happened in
this situation.
```

it seems to come from this line
https://github.com/cibernox/ember-power-select/blob/97621afa0284a2034a52684d80bbf766a981bb43/addon/components/power-select.js#L141

donaldwasserman:
This fixes #1094 

I just used the `get` function here, although i suspect there's probably a more elegant way to see if an object is `then-able`. 

I'm not sure how to write a test for this, but my fix doesn't break any existing tests and It Works On My Machine (and in my app) (TM).

### DIFF
--- a/addon/components/power-select.js
+++ b/addon/components/power-select.js
@@ -138,7 +138,7 @@ export default Component.extend({
       return null;
     },
     set(_, selected) {
-      if (selected && selected.then) {
+      if (selected && get(selected, 'then')) {
         this.get('_updateSelectedTask').perform(selected);
       } else {
         scheduleOnce('actions', this, this.updateSelection, selected);


cibernox:
It is clean enough for me.

cibernox:
Thanks!

PR #1102, opened 2018-04-25
Pre-render docs site at build time

ef4:
This PR enables Fastboot pre-rendering of the entire docs site using [prember[(https://github.com/ef4/prember). For URL discovery, it's using the newly-published [prember-crawler](https://github.com/ef4/prember-crawler)

I tested the `_redirects` on netlify to confirm it will work, and it does. By default, prember outputs an `_empty.html` file that contains your original (empty) index.html content, since your pre-rendered index.html file will have actual content in it.

### DIFF
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -2,13 +2,16 @@
 
 const EmberAddon = require('ember-cli/lib/broccoli/ember-addon');
 const VersionChecker = require('ember-cli-version-checker'); // eslint-disable-line
-
+const crawl = require('prember-crawler');
 
 module.exports = function(defaults) {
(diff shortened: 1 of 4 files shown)


cibernox:
Super interesting. I wasn't aware this could be so easy,

ef4:
Nice, I see the deployed site has fastboot content. 🎉 

```
$ curl -s http://ember-power-select.com | grep Powerful
    Powerful and customizable select component for
```

miguelcobain:
For reference, I also did this with ember-leaflet https://github.com/miguelcobain/ember-leaflet/commit/023683009a77d478fe9f4d2c2c20d24d59dab9da

Thank you!

PR #1117, opened 2018-05-23
fix IE11 placeholder workaround on ember 3.1

dwickern:
After upgrading from ember 2.18.2 to 3.1.1, the IE11 bug fixed by #355 reappeared. Evidently you need to return `undefined` now instead of `null`.


cibernox:
This has been fixed already, but thanks for the PR.

Issue #1114, opened 2018-05-14
Remove console.log statement from selectChoose helper


andreyfel:
This output looks odd in test logs.

PR #1119, opened 2018-05-24
Remove console.log from test helper

mydea:
Remove a console.log, which clutters the test output. See #1114 .


cibernox:
Thanks!

PR #1121, opened 2018-06-01
Remove requirement to provide a search field when using multiple select

alexspeller:
This field is not actually required, if you don't use it, you don't get the search filled in again when you press delete but that's not actually a problem if you dont' need that

### DIFF
--- a/addon/components/power-select-multiple/trigger.js
+++ b/addon/components/power-select-multiple/trigger.js
@@ -100,7 +100,6 @@ export default Component.extend({
               select.actions.search(lastSelection);
             } else {
               let searchField = this.get('searchField');
-              assert('`{{power-select-multiple}}` requires a `searchField` when the options are not strings to remove options using backspace', searchField);
               select.actions.search(get(lastSelection, searchField));
             }
             select.actions.open(e);


cibernox:
This assertion is becase 99% of the time if someone hits is because they forgot about the field, not because they don't want to delete with backspace.

I'd recommend that if you want to prevent deleting items with backspace you take an approach like the one in this twiddle: https://ember-twiddle.com/ad9bb34985ea8f072069dcea733d5eb9?openFiles=controllers.application.js%2C

**tl;dr:** Add an `onkeydown=fn` that does `if (e.keyCode === 8 && e.target.value === "") return false`

mayatron:
Thanks @cibernox for this tip. I used this to modify the behavior for `{{power-select-multiple}}` to allow backspace to completely remove the previous selection via an `onkeydown` function.

```
onKeyDown(api, e) {
  if (e.keyCode === 8 && isBlank(e.target.value)) {
    const lastSelection = api.selected[api.selected.length - 1];
    api.actions.select(api.selected.without(lastSelection));
    return false;
  }

  return true;
}
```

PR #1142, opened 2018-08-13
A11y Continued

RuslanZavacky:
We want to continue the effort in #953 on the latest EPS and updated a11y  based on the recent audit that we had and what Axe chrome plugin suggests around EPS.

I don't expect tests to pass straight away, some changes and improvements still required

### DIFF
--- a/addon/components/power-select.js
+++ b/addon/components/power-select.js
@@ -207,6 +207,11 @@ export default Component.extend({
     return this.get('searchEnabled') ? 'list' : 'none';
   }),
 
+  highlightedId: computed('publicAPI.highlighted', function() {
+    let { highlighted, results, uniqueId } = this.get('publicAPI');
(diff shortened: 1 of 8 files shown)


RuslanZavacky:
It actually seems that EPS master is quite good with a11y. The only problem Axe plugin report is when it opened, aria-controls refers to the wrong ID of a trigger. As the trigger is set just to ember generated ID, but in the dropdown, it uses a bit longer eg. `ember-power-select-trigger-ember53`. I'll probably create a separate ticket for that, to address.

RuslanZavacky:
Actually, after some voice over walkthrough. This version behaves better when the dropdown is opened. Eg. When you open dropdown for a master, it starts to pronounce *all* content that's inside. In this branch, every individual item will be pronounced separately, when selected. I'll research more, how to make this two versions work well.

PR #1144, opened 2018-08-16
guard against destroying hook when calling deactivate

cafreeman:
In Ember 3.2, a new assertion was added that prevents sending actions to objects that are already destroyed.

`ember-power-select` does this specifically with the `deactivate` action, and in our Ember 3.2 app this causing test failures because these actions are being sent after the component itself has been cleaned up.

Here are some examples of where this issue has come up elsewhere:
- https://github.com/emberjs/ember.js/issues/16820
- https://github.com/emberjs/ember.js/issues/16778

This PR prevents this error by wrapping calls to `deactivate` in `if (!this.isDestroying)`.

### DIFF
--- a/addon/components/power-select.js
+++ b/addon/components/power-select.js
@@ -377,7 +377,9 @@ export default Component.extend({
     },
 
     onTriggerBlur(_, event) {
-      this.send('deactivate');
+      if (!this.isDestroying) {
+        this.send('deactivate');
+      }
       let action = this.get('onblur');
       if (action) {
         action(this.get('publicAPI'), event);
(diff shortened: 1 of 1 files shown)


cafreeman:
@cibernox I'm not able to reproduce that one failing test in the `ember-2.16-lts` ember try scenario. Is it possible to re-run this node in Travis?

Fwiw, I can confirm that this does fix the issues in 3.2+ since I'm currently upgrading my work app now.

cibernox:
It was an spurious error. Merged! Thanks! I'll release a new patch version now

PR #1148, opened 2018-08-27
bugfix: "You attempted to access the length property ... of undefined"

marclundgren:
## ember error

```
You attempted to access the `length` property (of <(unknown):...).

Since Ember 3.1, this is usually fine as you no longer need to use `.get()`
to access computed properties. However, in this case, the object in question
is a special kind of Ember object (a proxy). Therefore, it is still necessary
to use `.get('length')` in this case.

If you encountered this error because of third-party code that you don't control,
there is more information at https://github.com/emberjs/ember.js/issues/16148, and
you can help us improve this error message by telling us more about what happened in
this situation.
```

## screenshots
screen shot 2018-08-27 at 12 52 14 pm
screen shot 2018-08-27 at 12 52 23 pm


abobwhite:
It looks like line 96 needs the same change, no?

abobwhite:
@cibernox Was line 96 not needed, then?

cibernox:
@abobwhite maybe it was. Did you experience a problem there?

abobwhite:
I am still experiencing problems even with both lines fixed, so it's hard to say. But line 96 has the same exact issue so I would imagine it would yield the same error. 

Here is the line with `onKeydown`:
`let lastSelection = select.selected[select.selected.length - 1];`

Issue #1147, opened 2018-08-21
Selected item not working


DenimTornado:
Template: 
`    {{#power-select selected='Choose category' verticalPosition='below' options=categories
                    onchange=(action "onchangeCat") as |category|}}
      <span>{{category.name}}</span>
      {{input value=category type="hidden" name="category"}}
    {{/power-select}}`

I don't see that 'selected' element.

broerse:
Just tested with ember-data 3.4 but it seems all versions above 3.1.1 break ember-power-select. See also https://github.com/emberjs/data/issues/5575

runspired:
@broerse unsure what power-select is doing but I don't think the issue is with ember-data

runspired:
I'd be skeptical of this: https://github.com/cibernox/ember-power-select/blob/master/addon/components/power-select.js#L542-L546

broerse:
@runspired no above ember-date 3.1.1 https://github.com/cibernox/ember-power-select/blob/master/addon/templates/components/power-select/trigger.hbs#L1 does not detect change on the new proxy object.

broerse:
@runspired see the test on https://ember-twiddle.com/8699e6c887c9eb2cb4344fb1d2551dbe?openFiles=templates.application.hbs%2C if you roleback to ember-data 3.1.1 it works fine.

broerse:
@runspired you can also see this failing now on https://bloggr.exmer.com . Press `Create` and select an author.

runspired:
@broerse so as I've said elsewhere, power-select should not rely on the proxy being swapped for a new instance.

It looks like from the twiddle that it works just fine in ember-data 3.4 for power-select, but the blockless version is broken. What's the implementation difference?

runspired:
Posting this comment here for visibility into what I suspect is going on: https://github.com/emberjs/data/pull/5584#issuecomment-417140955

broerse:
@cibernox I am not sure it can be fixed here because it seems to me that the ember-data change needs change on all levels. See Twiddle https://ember-twiddle.com/8699e6c887c9eb2cb4344fb1d2551dbe?openFiles=controllers.application.js%2C . There was a test created that nailed the problem but it was changed to test for other problem. See https://github.com/emberjs/data/pull/5584/commits/e01c74fff7502fc912a6db1cf5974dd4eb98039c

broerse:
[...]

PR #1151, opened 2018-09-07
Proxy object detection

jlami:
Test and possible fix for #1147

### DIFF
--- a/addon/components/power-select.js
+++ b/addon/components/power-select.js
@@ -9,6 +9,7 @@ import { DEBUG } from '@glimmer/env';
 import { isBlank } from '@ember/utils';
 import { isArray as isEmberArray } from '@ember/array';
 import ArrayProxy from '@ember/array/proxy';
+import ObjectProxy from '@ember/object/proxy';
 import layout from '../templates/components/power-select';
 import fallbackIfUndefined from '../utils/computed-fallback-if-undefined';
 import optionsMatcher from '../utils/computed-options-matcher';
@@ -143,7 +144,7 @@ export default Component.extend({
       return null;
     },
     set(_, selected) {
-      if (selected && get(selected, 'then')) {
+      if (selected && !(selected instanceof ObjectProxy) && get(selected, 'then')) {
         this.get('_updateSelectedTask').perform(selected);
       } else {
         scheduleOnce('actions', this, this.updateSelection, selected);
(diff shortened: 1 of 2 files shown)


cibernox:
@runspired Summoning you so you can assess this fix.

runspired:
@cibernox thinking it over

cibernox:
I restarted the failed job.

runspired:
My thoughts:

- I'm not sure this is generic enough although it should work for `ember-data` for now. There are other ways of constructing promise proxies than `ObjectProxy`, `ArrayProxy` and `PromiseProxyMixin`. I'm aware of many libraries that write their own simpler implementations, and I plan to write a simpler implementation for use in `ember-data` soon as well.

- This works towards solving the case where you encounter a PromiseProxy, but it seems likely there are still issues with generic proxies, which are even more likely to be stable than promise proxies

- I think the safest thing to do for PromiseProxies is to unwrap them. Unfortunately, the easiest way to do this is to observe a desired key (via observer or computed chain), which would be the best way to deal with generic proxies and other stable objects as well, but the level of abstraction here looks to prevent that at this time. That said, you could take an approach like the following to enable fixing `PromiseProxies` generically in the same manner as above in this PR.

```
function isPromiseProxy(potentialPromise) {
  if (typeof potentialPromise !== 'object' || potentialPromise === null || !potentialPromise.then) {
    return false;
  }

  // it is rare for a promiseProxy to not have one of `content` or `promise`
  // `unknownProperty` is the getter method that `Ember` looks for that allows an object to proxy
  if (potentialPromise.content || potentialPromise.promise || potentialPromise.unknownProperty) {
    return true;
  }

  return false;
}
```

Lastly, I'm working towards [eliminating](https://github.com/runspired/ember-data-experiments-mega-unstable) `PromiseProxy` and `PromiseArrayProxy` as a pattern used by `ember-data`, so there will come a happy future in which addon code does not be aware of these things so often. That said, the stable proxy and stable wrapper problem is very similar to the stable array problem, for which this library installs observers. It's likely that a little bit of refactoring would enable watching for changes in a more long-term vision manner.

cibernox:
In that case I think we can merge this. The test alone it's worth it, and if ember-data simplifies object proxies that should make things easier.

It might not work for other promise-proxy implementations, but this should cover most of the codebases and other people can resolve promises before passing them to EPS.

PR #1155, opened 2018-09-12
Only load polyfill when not in FastBoot environment

### LINKED ISSUE #1154: Options component broken in fastboot
https://github.com/cibernox/ember-power-select/blob/7ffe3f0a3d2fdc7132b3bf991245d93610f0ef7c/addon/components/power-select/options.js#L6-L23

The above polyfill results in an infinite rendering invalidation in a fastboot environment.

The basic error:
`TypeError: Cannot read property 'prototype' of undefined`

I think we could just wrap it in a:
```
if(typeof FastBoot === undefined){
    ... polyfill ...
}
```

nickschot:
fixes #1154

### DIFF
--- a/addon/components/power-select/options.js
+++ b/addon/components/power-select/options.js
@@ -3,24 +3,26 @@ import { computed } from '@ember/object';
 import layout from '../../templates/components/power-select/options';
 
 const isTouchDevice = (!!window && 'ontouchstart' in window);
-(function(ElementProto) {
-  if (typeof ElementProto.matches !== 'function') {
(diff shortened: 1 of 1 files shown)


cibernox:
Makes a lot of sense!

cibernox:
Published a new patch version for this. Great catch!

nickschot:
I probably ran into it because I use an extended options component which I pass into power-select.

cibernox:
That can also cause the component to be eagerly imported and make the polyfill run. Either way, you fixed it :D

PR #1162, opened 2018-09-26
use .set() to avoid assertion failure under Ember 3.4

seansellek:
While updating Ember to 3.4.2, we encountered issues with ember-power-select throwing:

screen shot 2018-09-26 at 10 58 49 am

This pr changes two instances where `openingEvent` is set to use `Ember.set` as opposed to `=`.

### DIFF
--- a/addon/components/power-select.js
+++ b/addon/components/power-select.js
@@ -228,7 +228,7 @@ export default Component.extend({
         return false;
       }
       if (e) {
-        this.openingEvent = e;
+        this.set('openingEvent', e);
         if (e.type === 'keydown' && (e.keyCode === 38 || e.keyCode === 40)) {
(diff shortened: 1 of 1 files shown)


cibernox:
That's weird. That properly is not tracked in any template, so it seems to me that ember should allow to use a regular assignment there. AFAIK `set` is only required for properties that are tracked by templates or computed properties.

This seems fine, and I'll merge it, but I'll ask if this warning is intentional.

seansellek:
I don't have an explanation either. If it helps, I'm using ember-bootstrap-power-select as well.

cibernox:
Merged! I'll release tomorrow, you can use master in the meantime

seansellek:
Awesome! Thank you =)

PR #1165, opened 2018-10-02
Add animationEnabled to be passed down to ember-basic-dropdown

neeleshsaxena:
Need to pass in `animationEnabled` property through ember-power-select, so that it can be used by ember-basic-dropdown.

Been running into ember run loop errors related to animating dropdown when the component has already been destroyed, as it's scheduled afterRender here - https://github.com/cibernox/ember-basic-dropdown/blob/master/addon/components/basic-dropdown/content.js#L183

Had submitted a similar PR for this property to be passed in through ember-power-select-typeahead, which has been merged recently. 
https://github.com/cibernox/ember-power-select-typeahead/pull/77


neeleshsaxena:
jenkins build

neeleshsaxena:
👍

PR #1166, opened 2018-10-04
Add highlightOnHoverEnabled to be used in options.js

cah-danmonroe:
When false, will disable automatically highlighting an option
when hovered over with the mouse. This will prevent unintentionally
selecting a new value when tabbing through a form with an EPS field
while the user's mouse is where the list will appear when the EPS
gains focus, it is set to automatically open onFocus, and then closes 
when it loses focus.

### DIFF
--- a/addon/components/power-select.js
+++ b/addon/components/power-select.js
@@ -86,7 +86,7 @@ export default Component.extend({
   closeOnSelect: fallbackIfUndefined(true),
   defaultHighlighted: fallbackIfUndefined(defaultHighlighted),
   typeAheadMatcher: fallbackIfUndefined(defaultTypeAheadMatcher),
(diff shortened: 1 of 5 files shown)


addon/components/power-select.js
```diff
@@ -86,6 +86,7 @@ export default Component.extend({
   closeOnSelect: fallbackIfUndefined(true),
   defaultHighlighted: fallbackIfUndefined(defaultHighlighted),
   typeAheadMatcher: fallbackIfUndefined(defaultTypeAheadMatcher),
+  highlightOnHoverEnabled: fallbackIfUndefined(true),
```
cibernox: Can you just rename this to `highlightOnHover` (without the `enabled` suffix). Seems redundant.

cibernox:
I only have one small request, then I'll merge it.

Issue #1141, opened 2018-07-31
Default click behavior breaks in Chrome when upgrading from Ember 2.18 to 3.1


goo32:
The default behavior in Ember 2.18 when clicking on the trigger: the search input is displayed, focused and ready to start typing.

After upgrading to Ember 3.1 and only in Chrome (I've only tested in Chrome and IE 11) the input does not focus so you can't just start typing.

You can see this in the first example in the [How to use it](https://ember-power-select.com/docs/how-to-use-it) docs.

nickschot:
I think I'm also seeing this issue (ember 3.4 e-p-s 2.0.8). Hope to be able to debug it later in the week.

cibernox:
I missed this initially. I can also reproduce the bug. I'll try to find a moment to debug it.

PR #1167, opened 2018-10-11
fixed search not focusing

robclancy:
fixes https://github.com/cibernox/ember-power-select/issues/1141

Not sure the real issue but we need this to work, it's terrible UX without it.


knownasilya:
Would love to see this merged. @robclancy looks like you need to rebase before it can be merged.

cibernox:
If this is rebased I think we can safely merge it.

robclancy:
Isn't it fixed with https://github.com/cibernox/ember-power-select/commit/6bc13ba270bf49ec3d386ee2666d445d0afa946b?

cibernox:
@robclancy I believe it should. Can you virify?

PR #1169, opened 2018-10-18
Change Slack for Discord

esbanarango:
#### Before
screen recording 2018-10-18 at 09 48 am

#### After
screen recording 2018-10-18 at 09 44 am


cibernox:
Good catch!

PR #1176, opened 2018-11-15
close -> clone

jfdnc:
'clone' is the intended word in the context, and the spelling error 'close' may cause some confusion to the reader


cibernox:
Good catch!

PR #1177, opened 2018-11-15
Improve documentation

Caltor:
I have improved the first paragraph to use the active-voice rather than the passive-voice and documented (including an example) now to use a computed property for the `selected` parameter of this component. This PR also includes a fix for the existing build error and updates the contributing.md to use yarn instead of npm and bower.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,8 +14,7 @@ The documentation website lives under `tests/dummy`. It is a simple
 ember app. You can run the website locally:
 
 ```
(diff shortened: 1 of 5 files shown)


cibernox:
Thanks!

Caltor:
NP happy to contribute even in a small way to this excellent add-in. Thanks for all your hard work. I tried adding an example using `find-by` helper but it needs ember-composable-helpers and it all went pear-shaped. A job for another day maybe...

PR #1188, opened 2018-12-03
Custom search field example

samdemaeyer:
## ❗️DO NOT MERGE ❗️
Is this a bug? Or are we doing something wrong?!
If I type into the new input field (Custom Component) it works, however, when we do it more than once, we receive this error:
```
"Assertion Failed: {{power-select}} If you want the default filtering to work on options that are not plain strings, you need to provide `searchField`"
```
sgneaxps0i


samdemaeyer:
@cibernox, thanks for the quick response. The problem we are having is that this is a custom search field, using the `triggerComponent`. So there should not be a need to specify the `searchField`.
Thanks to @eugeniodepalo, we did figure out what was wrong. I was `setting` the options in the custom `search=(perform searchTask)`. This should not be set in there, but the task should return the options.
So no need for `searchField` nor `options=options`. 👇 
```hbs
<!-- tests/dummy/app/templates/custom-search.hbs -->
{{#power-select
  allowClear=allowClear
  onchange=(action "select")
  placeholder="Type to search..."
  beforeOptionsComponent=null
  search=(perform searchTask)
  searchEnabled=false
  selected=selected
  triggerComponent="search-trigger"
as |obj|}}
  {{obj.text}}
{{/power-select}}
```
```js
  searchTask: task(function*(query) {
    yield timeout(300);
    const searchOptions = [
      { text: `Suggestion: ${query} 1` },
      { text: `Suggestion: ${query} 2` },
      { text: `Suggestion: ${query} 3` },
      { text: `Suggestion: ${query} 4` },
      { text: `Suggestion: ${query} 5` }
    ];
    // DO NOT SET IT HERE, RETURN THEM INSTEAD
    // this.set('options', searchOptions); 
    return searchOptions;
  }),
```
This actually solved the problem! Thanks for your help @eugeniodepalo

PR #1201, opened 2019-02-03
Prevent scroll only if target is body

samdemaeyer:
~We should only prevent the scroll when the target is the body. I have an issue that my target is an input field with a custom `triggerComponent`. The `e.preventDefault()` is preventing me from typing in a space 😅~
We should not handle the spacebar if the node is a `textarea` or an `input` field. I have an issue when my target is an input field with a custom `triggerComponent`. 

**Before:**
kapture 2019-02-03 at 11 48 43

**After:**
kapture 2019-02-03 at 11 50 00

### DIFF
--- a/addon/components/power-select.js
+++ b/addon/components/power-select.js
@@ -657,11 +657,12 @@ export default Component.extend({
   },
 
   _handleKeySpace(e) {
+    if (['TEXTAREA', 'INPUT'].includes(e.target.nodeName)) {
+      return false;
+    }
     let publicAPI = this.get('publicAPI');
     if (publicAPI.isOpen && publicAPI.highlighted !== undefined) {
-      if (e.target === document.body) {
-        e.preventDefault(); // Prevents scrolling of the page.
-      }
+      e.preventDefault(); // Prevents scrolling of the page.
       publicAPI.actions.choose(publicAPI.highlighted, e);
       return false;
     }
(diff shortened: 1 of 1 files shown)


addon/components/power-select.js
```diff
@@ -659,7 +659,9 @@ export default Component.extend({
   _handleKeySpace(e) {
     let publicAPI = this.get('publicAPI');
     if (publicAPI.isOpen && publicAPI.highlighted !== undefined) {
-      e.preventDefault(); // Prevents scrolling of the page.
+      if (e.target === document.body) {
+        e.preventDefault(); // Prevents scrolling of the page.
```
samdemaeyer: @cibernox, I see you have covered this `preventDefault()` in your test [keyboard-control-test.js:174](https://github.com/cibernox/ember-power-select/blob/master/tests/integration/components/power-select/keyboard-control-test.js#L174)
I was thinking .. I read somewhere that scrolling happens on the body. I'm not sure if that is correct though. Maybe we can detect if it's an elemnt where you can type something. Then we can actually test it. Otherwise, should we remove the test? This will never be `true` => `e.target === document.body`  in the test I think
```js
if (['TEXTAREA', 'INPUT'].includes(e.target.nodename)) {
  e.preventDefault(); 
}
```
What do you think?

samdemaeyer: Actually .. come to think of it .. we should not handle the spacebar at all when we are in an input field or a text area .. 😅

PR #1204, opened 2019-02-19
[WIP] Update to octane

cibernox:


sly7-7:
Currently installing ember-power-select in an app built with the octane blueprints fails with `ENOENT: no such file or directory, mkdir 'app/styles'`, I think it's when eps addon try to add its scss import in app.scss. Is this something to do in this pr ?

cibernox:
This PR hasn't been merged so no, it can't be. Could it be because the octane blueprint uses Module Unification? I believe that in MU there shouldn't be /app folder.

sly7-7:
I don't know, and I tested this morning installing EPS in a fresh new app using octane blueprints, and it worked.

simonihmig:
@cibernox out of curiosity (planning similar changes in https://github.com/kaliber5/ember-bootstrap/issues/780), how do you plan to change the support for Ember versions? 

It *has* to be a breaking change AFAIK, as `ember-angle-bracket-invocation-polyfill` supports only 2.12+, right? Or are you going to be so bold and support only 3.4+ (which has native angle bracket support and is the only active LTS anyway)?

cibernox:
@simonhmig yeah, most likely I'll support only 3.X as I'll be using the very latest and greatest features, which probably won't be pollyfillable. 

But even if I didn't there will ha be breaking changes that will requiere a version bump.

Issue #1057, opened 2018-02-05
power-select-multiple component tag rendering twice


Techn1x:
I'm using these versions
```
ember-power-select v1.10.4
ember-power-select-blockless v0.3.1
ember-power-select-with-create
```

I have the following in a template
```handlebars
{{#power-select-multiple
  class="flexbox-row-full margin-10-left"
  options=options
  selected=selected
  onchange=onChangeAction
  matcher=matcher
  as |id| }}
  ...
{{/power-select-multiple}}
```

Looking at my page, it renders this;
![capture](https://user-images.githubusercontent.com/1049837/35784572-d08060c8-0a64-11e8-8111-910df3ee07b6.PNG)
As you can see, it renders the same tag twice (but nested). Similarly, if I set `tagName="span"`, then both of those tags become spans.

The API reference for `class` says that it shouldn't even do anything unless tagName is specified
```
The CSS class of the component. The component by default is tagless, so this won't take any effect unless the user specifies `tagName=div` too.
```
http://ember-power-select.com/docs/api-reference

What's going on here? Is this normal?
Thanks

gwak:
I seeing the same behaviour... with latest version.
Actually even on the documentation website, you can see it.

Techn1x:
It looks like 2.2.0 fixes the `tagName` issue (ie if not specified, then the element should be tagless) https://github.com/cibernox/ember-power-select/pull/1185

However when specifying a `tagName`, we still get the double element render.

Techn1x:
(workaround for us is to use EPS 2.2.0+ and always use EPS as tagless, and just surround the EPS element with the div we want)

PR #1206, opened 2019-03-01
Fix Issue 1057 - component tag rendering twice

mpirio:
cf https://github.com/cibernox/ember-power-select/issues/1057

When we set a `tagName` on a `power-select-multiple`, this tag is rendering twice because `tagName` is set for `power-select-multiple` but it is transmit to `power-select` also. I removed `class` too.

improvment idea: Perhaps could we set a default "div" `tagName` when a `class` is setted?

### DIFF
--- a/addon/templates/components/power-select-multiple.hbs
+++ b/addon/templates/components/power-select-multiple.hbs
@@ -11,7 +11,6 @@
     beforeOptionsComponent=beforeOptionsComponent
     buildSelection=(action "buildSelection")
     calculatePosition=calculatePosition
-    class=class
     closeOnSelect=closeOnSelect
     defaultHighlighted=defaultHighlighted
     destination=destination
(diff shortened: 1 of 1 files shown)


cibernox:
If you only pass `class` right now, doesn't that give you a  single div already?

mpirio:
@cibernox I tested with last version 2.2.2: no 'div'  is added automatically if we define a `class`.

mpirio:
@cibernox what do you think about this PR? ;) Thanks

cibernox:
Sorry, I forgot about this. Checking

cibernox:
I have an alternative solution that I think it's more aligned with the future in which components are tagless by default, and it's not a breaking change today.

The problem is that since `tagName` has a meaning for components, is applied to both elements, which only the internal one should have it. The multiple should only be a wrapper. 

I suggest that the way to customize the tag of a multiple-select will be now `htmlTag`. Internally you pass `tagName=(or tagName htmlTag)` to EPS. That makes it backwards compatible. And you don't need to touch `class`.

PR #1218, opened 2019-04-08
Add option to not close select on tab keypress

kstinson14:
I would like to add an optional property to the `power-select` component to allow the dropdown to stay open on a "Tab" keypress.  You currently can pass the `onkeydown` method to the component and do some customization that way.  You can then return `false` in this method to prevent the default closing behavior from occurring.  But my use case involves tabbing through / focusing content within the dropdown portion of the select.  If I were to return false in `onkeydown`, the default browser focus behavior would also be prevented.  Creating this property allows the `power-select's` default tab behavior to be avoided, while allowing the browser's default tab behavior to be preserved.

### DIFF
--- a/addon/components/power-select.js
+++ b/addon/components/power-select.js
@@ -84,6 +84,7 @@ export default Component.extend({
   noMatchesMessage: fallbackIfUndefined('No results found'),
   searchMessage: fallbackIfUndefined('Type to search'),
   closeOnSelect: fallbackIfUndefined(true),
+  closeOnTab: fallbackIfUndefined(true),
   defaultHighlighted: fallbackIfUndefined(defaultHighlighted),
   typeAheadMatcher: fallbackIfUndefined(defaultTypeAheadMatcher),
   highlightOnHover: fallbackIfUndefined(true),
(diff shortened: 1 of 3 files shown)


cibernox:
@kstinson14 I'm not sure about this:
> If I were to return false in onkeydown, the default browser focus behavior would also be prevented

AFAIK, returning `false` from `onkeydown` does not prevent any native browser behavior whatsoever, only the default behavior of the select (unless you call `event.preventDefault()`).

Can you check that and explain a bit more why returning false doesn't work for your use case?

kstinson14:
@cibernox Sorry, I should've included this in the original post: I have an [ember-twiddle](https://ember-twiddle.com/5e4ea55a266749387f7f2c1c685ffbf6?openFiles=templates.application.hbs%2C) here that better shows what I mean.  If you have an `onkeydown` handler on an input that returns false, it prevents the focus from changing to the next element in the tab order.  You can see this in the second input / button pair in the twiddle.  

I also put a small `power-select` example in there that displays the problem more explicitly.  When you click on the trigger to open the component, there are buttons in each of the options.  But they can't be tabbed to, because the input's onkeydown handler returns false.  If you alter the `power-select` component to have `searchEnabled=false`, the buttons become tab-able, since the input's onkeydown event won't get in the way.

kstinson14:
@cibernox Realized I didn't add the attribute to `power-select-multiple`, should be fixed now.

rwwagner90:
@kstinson14 I realize this is super old, but could you please rebase and fix conflicts?

mkszepp:
closing this super old PR as it's not anymore mergable & no feedback since years.. feel free to repo if you still need /want this feature

PR #1222, opened 2019-04-28
fix disabled options on touch devices

swoga:
This PR fixes a problem which allows to select disabled options on touch devices.

### DIFF
--- a/addon/components/power-select/options.js
+++ b/addon/components/power-select/options.js
@@ -2,7 +2,6 @@ import Component from '@ember/component';
 import { computed } from '@ember/object';
 import layout from '../../templates/components/power-select/options';
 
-const isTouchDevice = (!!window && 'ontouchstart' in window);
 if(typeof FastBoot === 'undefined'){
   (function(ElementProto) {
     if (typeof ElementProto.matches !== 'function') {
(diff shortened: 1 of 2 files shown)


cibernox:
Good catch! Can you add a test for this?

swoga:
I have added a test for disabled options.
I needed to remove isTouchDevice, because when testing on desktop it is always set to false and I found no trivial way to overwrite it for testing.
I don't see any downside adding the event, even when there is no touch support.
The touch control tests also were only successful, because tap() fires onclick, but it didn't test the logic behind the touch events.

cibernox:
@swoga you shouldn't have to remove that, you can run chrome with a specific flag to make it think it's on a touch device. Tests in CI run with that flag on:
https://github.com/cibernox/ember-power-select/blob/master/testem.js#L11

swoga:
Thanks for the hint! I didn't know this parameter and simply ran the tests inside the browser.
Using `ember test` runs the tests as expected.
I have reverted the commit.

cibernox:
Great, I'll merge and release a new version as soon as travis is green!

cibernox:
@swoga thank you for the PR. I've release 2.3.2 with this fix 🎉

PR #1227, opened 2019-05-31
Update to Ember & Ember-cli 3.10

cibernox:


gorner:
Not sure if this was intentional / expected for a minor release, but as a result of this blueprint update this add-on has apparently dropped support for Node 6.x. I understand that Node 6 is now EOL, but dropping support for old versions of Node has normally been considered a breaking change for Ember addons.

cibernox:
@gorner Since it was EOL, I didn't consider it a breaking change. FWIW, since nothing in the code has really changed, node 6 should continue to work, it's just there because it's part of the blueprint when you update ember.

I'm happy to add it back to the list of supported engines of the package.json if that makes sense. 

I certainly don't want to release 3.0 because of this, as there is no API changes.

gorner:
@cibernox Up to you. It's probably time we update our app to use Node 8+ for CI anyways.

JibranKalia:
Yarn doesn't let us install the package: 
`error ember-power-select@2.3.3: The engine "node" is incompatible with this module. Expected version "8.* || >= 10.*". Got "6.14.4"`

cibernox:
@JibranKalia I released 2.3.4 adding back node 6 support. Although since node 6 is officially dead (It was released in April 2016) you should consider updating

JibranKalia:
You are absolutely right. We need to update. Thanks for doing this @cibernox

PR #1233, opened 2019-06-09
V3.0 beta

cibernox:
Preview can be seen in https://v3-0-beta.ember-power-select.com


cibernox:
Now that 3.11 is released, this is ready to be tested

PR #1235, opened 2019-06-24
guard against destroying hook when calling activate

AmilKey:
Assertion Failed: Attempted to call .send() with the action 'activate' on the destroyed object <component:power-select::ember479>


AmilKey:
@cibernox please take a look at

cibernox:
This should be fixed in master now.

PR #1266, opened 2019-08-23
searchEnabled defaults to false as Breaking change

sly7-7:
Not sure it's worth it, but I would find it useful to have this mentioned


cibernox:
Good catch. Thanks!

PR #1276, opened 2019-09-16
typo: Remove this._super to use native super

dcyriller:
This was the only usage of `this._super` in a native class. A last occurence is remaining in the tests but in an Object: `Component.extend({})`.

#### Context
As explained [here](https://blog.emberjs.com/2019/01/26/emberjs-native-class-update-2019-edition.html), native `super` should be used over `this._super` in native classes.


cibernox:
Good catch! Thanks

PR #1277, opened 2019-09-16
Bump eslint-plugin-ember to v7

dcyriller:
https://github.com/cibernox/ember-power-select/pull/1276 would have been caught by the recently added [no-ember-super-in-es-classes rule](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-ember-super-in-es-classes.md)

@cibernox is there any reason why you don't use something like dependabot? If you'd like I could open a PR to add the config (which I think would turn it on).

Thank you for the great work on this addon!

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,7 +81,7 @@
     "ember-source-channel-url": "^1.1.0",
     "ember-try": "^1.1.0",
     "eslint": "^5.16.0",
-    "eslint-plugin-ember": "^6.3.0",
+    "eslint-plugin-ember": "^7.0.0",
     "eslint-plugin-node": "^9.0.1",
     "liquid-fire": "0.30.0",
     "loader.js": "^4.7.0",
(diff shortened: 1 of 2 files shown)


dcyriller:
The CI is expected to fail on linting while #1276 has not been merged 🙂

cibernox:
@dcyriller last time I tried using dependabot in a project it was super noisy. Is there a way to limit the number of updates to a max of X a week?

dcyriller:
Rebased!

> last time I tried using dependabot in a project it was super noisy

Agreed. For my small-sized projects I configure it:
- to run once a month
- to squash and merge, not to have too many commits on master

cibernox:
@dcyriller if you can configure to run like twice a month i'd totally love that! 😍

dcyriller:
ahah it is actually either weekly or monthly. Unfortuntately, twice a month doesn't look available to me: https://dependabot.com/docs/config-file/#update_schedule-required

cibernox:
@dcyriller let's start with a weekly approach. I try to review issues and PRs every Friday anyway.

PR #1278, opened 2019-09-16
chore: Add dependabot config

dcyriller:
As discussed in https://github.com/cibernox/ember-power-select/pull/1277, it will run on a weekly basis.

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -1,0 +1,11 @@
+version: 1
+update_configs:
+  - package_manager: "javascript"
+    directory: "/"
+    update_schedule: "weekly"
+    default_labels:
+      - "dependencies"
+    automerged_updates:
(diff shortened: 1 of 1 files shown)


dcyriller:
Here you go @cibernox!

This config should also auto-merge in-range updates. To enable that feature, you'll have to visit https://app.dependabot.com/accounts/cibernox/settings and tick the box `Allow auto-merging to be enabled on projects`.

To enable squash and merge for dependabot PRs, you'll have to untick the box `Create a merge commit if a PR is merged by Dependabot`.

And it should be good!

One thing to control is that dependabot do not narrow the direct `dependencies` range (like `ember-basic-dropdown`). It should be okay, but you can check it in the first PRs!

cibernox:
thanks! This should make maintenance easier.

PR #1293, opened 2019-11-19
Make `results` and `resultCount` computed properties

cibernox:
If those properties are just getters, they get called pretty often: `results`
was called 4 times for every keystroke in the search box, and `resultsCount` 
twice.

Since filtering iterates over every element of the list (potentially a hundreds 
of elements), calling it so ofter was possibly inefficient. 
Counting the results also is a function that iterates all options (and suboptions 
if there are groups) so it better to cache the result.

From a first swipe of the component, those two seem the only getters worth caching

### DIFF
--- a/addon/components/power-select.js
+++ b/addon/components/power-select.js
@@ -2,6 +2,7 @@
 import Component from '@glimmer/component';
 import { tracked } from '@glimmer/tracking';
 import { action, get, computed } from '@ember/object';
+import { dependentKeyCompat } from '@ember/object/compat';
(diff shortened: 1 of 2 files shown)


cibernox:
Apparently `dependentKeyCompat` may not be supported for a long time, so not sure if it's worth using it.

cibernox:
This is closed and #1295 is created instead, in which the filtered results are cached, as that is really the only scenario in which the results getter can get a little bit expensive.

PR #1305, opened 2019-12-04
Ensure multiple select <ul> has only <li> children

mydea:
ember-a11y-testing (or more specifically, axe) complains that a `<ul>` has to have `<li>` children only.
This PR slightly changes the DOM structure of the multi-select trigger to accomodate that.

I have tried it, and it should look/feel/behave the same!

### DIFF
--- a/addon/components/power-select-multiple/trigger.hbs
+++ b/addon/components/power-select-multiple/trigger.hbs
@@ -26,34 +26,38 @@
     </li>
   {{else}}
     {{#if (and @placeholder (not @searchEnabled))}}
-      <span class="ember-power-select-placeholder">{{@placeholder}}</span>
+      <li>
+        <span class="ember-power-select-placeholder">{{@placeholder}}</span>
+      </li>
     {{/if}}
   {{/each}}
   {{#if @searchEnabled}}
-    <input
-      type="search"
-      role="combobox"
-      class="ember-power-select-trigger-multiple-input"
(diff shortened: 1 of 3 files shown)


cibernox:
I'll pull the PR and investigate if there's a less breaking way of achieving the same thing, but I agree with the idea. I want this to be part of v4

mydea:
Awesome! Yeah, I have also looked a bit around. Another idea would be to simply replace the `<ul>` and `<li>` with `<div>` - also not sure if that would otherwise negatively impact the accessibility...!

MarcoZehe:
Can this be moved along, too?

rwwagner90 (approved):

rwwagner90:
@mydea is there any possibility of this being a breaking change? Especially if people were styling assuming the `li` was not there?

mydea:
It's _possible_, yes! But I would say rather unlikely, e.g. if you relied on `ul > .ember-power-select-placeholder` it would break, I guess.

I guess the biggest potentially breaking change is that the container (`.ember-power-select-multiple-options`) has `display: flex`, which _could_ affect styling. Not sure what guarantees there are about this internal markup!

gvdp:
It was a breaking change for our styling, because the precompiled css is not updated: https://github.com/cibernox/ember-power-select/issues/1586

Also broke a unit test which checked for the presence of a single `<li` element for the selected option, but there are now 2. 

Not sure this actually improves a11y, it just satisfies a linting rule. The `<input` field should probably exist outside of the `<ul` list, no?

PR #1306, opened 2019-12-04
Fix a11y issues with inputs

mydea:
This PR fixes (or at least allows to fix) lingering accessibility issues reported by axe (via ember-a11y-testing):

* Pass `aria-label` & `aria-labelled-by` to all search inputs
* Remove `role="combobox"` from search input (See https://github.com/dequelabs/axe-core/issues/1580)

It builds on top of #1305 - I can change that, if required.

With these changes, I could get axe to pass - as long as I added a label, like this:

```hbs
<label id='my-label'>Label here</label>
<PowerSelect
  @ariaLabelledBy='my-label'
  ...
>
```

IMHO it would be nice to add this somewhere to the docs, maybe to the cookbook? But that is a separate issue :)

### DIFF
--- a/addon/components/power-select-multiple/trigger.hbs
+++ b/addon/components/power-select-multiple/trigger.hbs
@@ -47,7 +47,7 @@
         aria-labelledby={{@ariaLabelledBy}}
         aria-label={{@ariaLabel}}
         value={{@select.searchText}}
-        aria-controls={{@listboxId}}
+        aria-controls={{if @select.isOpen @listboxId}}
         style={{this.triggerMultipleInputStyle}}
         placeholder={{this.maybePlaceholder}}
         disabled={{@select.disabled}}
--- a/addon/components/power-select-multiple/trigger.hbs
+++ b/addon/components/power-select-multiple/trigger.hbs
@@ -35,7 +35,6 @@
     <li>
       <input
         type="search"
-        role="combobox"
         class="ember-power-select-trigger-multiple-input"
         aria-activedescendant={{if @select.isOpen @ariaActiveDescendant}}
         aria-haspopup="listbox"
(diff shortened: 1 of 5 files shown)


mydea:
Anything I can do to help move this along (or change)? Would love to see this land :)

MarcoZehe:
I'd also be interested in seeing this merged. Ghost uses this component and has some serious issues with keyboard navigation when the PowerSelect is open, and labeling. @cibernox can we work together to move this along? Also with #1305 .

mydea:
I have updated this a bit, also adding `id`, `aria-label`/`aria-labelledby` to all "empty" listboxes (when showing the search or empty message).

backspace:
Thanks for opening this, I skipped an accessibility audit in tests that use Power Select for now because it seemed like it would involve too many template overrides. I know @cibernox is always busy but let us know if there’s any way we can help this move along 💞

rwwagner90:
@mydea would you be able to rebase this and fix the conflicts please?

mydea:
sorry, took some time, but managed to rebase it.

FYI, there are still some major accessibility issues, but they are harder to fix... The main problem I see is that there are nested interactive elements in the multi select (the trigger is interactive, and then inside of it there are both the "x" remove buttons as well as the search input).

IMHO the way to "solve" this is to refactor the multi-select to also have the search in the dropdown, similar to the single select. But this is of course a larger change. I wrote some thoughts together here: https://github.com/cibernox/ember-power-select/issues/1539

rwwagner90:
@mydea I merged #1305 so unfortunately you will need to rebase this again. Sorry! Do we need to label any of this stuff as breaking changes?

mydea:
I rebased, and I _think_ this should be backwards compatible, as it only adds/remove some attributes, but doesn't change any structure.

rwwagner90 (approved):

PR #1315, opened 2019-12-26
Document how to skip styles

josemarluedke:
I'm working on an addon that will wrap ember-power-select and provide its own styles using tailwind plugins. In that respect, I won't be using sass nor less, and I don't want the pre-compiled CSS.


cibernox:
You can already not use the styles if you pass `theme: false` in the config. But I agree that it's not mentioned in the docs. Does that option work for you? If so you only have to document it.

josemarluedke:
@cibernox That seems to work as expected. I updated the PR to just add documentation around this. Let me know if you have any other concerns.

cibernox:
Thanks!

PR #1330, opened 2020-02-09
update readme with minimum required ember version for 4.0

cibernox:


cibernox:
@chancancode It is indeed fixed

PR #1331, opened 2020-02-12
Pass splattributes correctly, fixes #1203

### LINKED ISSUE #1203: Unable to pass arbitrary attributes into the trigger
I need to pass a `data-test-...` attribute into the trigger. 

Looks like this is impossible to do.

Please pass splattributes (or some pre-Octane equivalent) into the trigger.

Or give the wrapper component a tag.

lolmaus:
Passing splattributes into `<BasicDropdown>` does nothing because `<BasicDropdown>` is tagless.

This causes #1203: it is impossible to apply attributes to the trigger component.

This simple change fixes the problem.


lolmaus:
A bunch of tests are failing with `'obj.searchTask)' is not a valid attribute name.`. 

I'm not sure if my change is responsible for those failures.

@cibernox, please advise.

andreyfel:
@lolmaus you might want to rebase your PR since #1377 has been merged to master. It should pass after that.

lolmaus:
Done. Thx, comrade!

lolmaus:
PR rebased.

CI seems to be [paused](https://github.com/cibernox/ember-power-select/actions/runs/924603442) with:

> This workflow is awaiting approval from a maintainer in #1331

@cibernox This tiny PR is now over a year old, we and other projects depend on it. Can this please be merged and released?

cibernox:
Running CI. I'll merge if green. 
This can _technically_ be breaking because someone might have extended BasicDropdown to not be tagless, but it is extremely unlikely.

cibernox:
Thanks! Sorry this sat here for so so long.

nickschot:
Big bump, but just to add for future travelers: this change is breaking for the case where the power-select/basic-dropdown is rendered in place as the splattributes were passed to the wrapping element of the basic-dropdown in that case meaning 4.1.6 -> 4.1.7 is a breaking change. Unfortunately the change did not show up in the changelog, so it was kinda hard to track this down 😬 .

gvdp:
> Big bump, but just to add for future travelers: this change is breaking for the case where the power-select/basic-dropdown is rendered in place as the splattributes were passed to the wrapping element of the basic-dropdown in that case meaning 4.1.6 -> 4.1.7 is a breaking change. Unfortunately the change did not show up in the changelog, so it was kinda hard to track this down 😬 .

I concur, our styling broke after upgrading to 4.1.7 and this was the reason, pretty hard to figure out :-)

Issue #1334, opened 2020-02-20
[v3.0.6] Autocloses if content overlays trigger


mehulkar:
I'm trying to achieve a UI where the dropdown opens above the trigger. I do this by applying a `transform: translateY()` onto the content element.

```hbs
<PowerSelect
  @options={{this.options}}
  @selected={{this.selectedOption}}
  @onChange={{this.setSelection}} as |opt|>
  {{opt}}
</PowerSelect>
```

```scss
.ember-power-select-dropdown.ember-basic-dropdown-content {
  transform: translateY(-32px);
}
```

When I click on the trigger, the dropdown closes as soon as it's opened.

## Investigation

What's happening is that when I click on the trigger, when the content is on top of the trigger, the click event is dispatched by the `body` element (I'm not sure why?). That means that the callback decides to close the dropdown, since it's [not a click on the trigger][1]. If the content is *not* on top of the trigger, the `click` event is dispatched by the trigger element, and it effectively ignores the event.

## Possible solutions

1. If I pass the `rootEventType` to `mouseup` to EBD, I get the behavior I want. The semantics between click and mouseup are different, but I believe it's correct behavior for my usecase. **However EPS does not expose this to the end user**, so it would need to be added as public API and [passed into EBD here][2].
2. Another solution is to figure out *why* the `click` event is being dispatched on `body`. I'm guessing it's because the `mousedown` is on the trigger and `mouseup` is on the content, so the closest ancestor is the `body`? I looked into the `stopPropagation` API as a possible solution for this, but I am not sure it's the right solution.

Here's the UI I'm trying to get:

![bugreport](https://user-images.githubusercontent.com/490968/74890311-84d44800-5338-11ea-97e9-5dca492f8cf7.gif)

One issue I notice with changing `rootEventType` to `mouseup` though is that if I don't move my mouse at all and click again, it doesn't go away. I haven't figured out what the issue is there.

[1]: https://github.com/cibernox/ember-basic-dropdown/blob/v2.0.13/addon/components/basic-dropdown-content.js#L123-L136
[2]: https://github.com/cibernox/ember-power-select/blob/v3.0.6/addon/templates/components/power-select.hbs#L2-L15

mayatron:
Seeing the same issue here.

Possibly related? https://github.com/cibernox/ember-basic-dropdown/issues/508

mayatron:
I can reproduce this issue on EPS `4.0.0-beta.6`.

[...]

PR #1333, opened 2020-02-19
Fix menu autoclose when content is rendered over trigger.

praveendinesh6:
Fixes the issue - https://github.com/cibernox/ember-power-select/issues/1282

Issue: When moving the dropdown content to cover the trigger element for design reasons using CSS, the content prematurely closes on each click.

Further discussion on this issue [here](https://github.com/cibernox/ember-power-select/issues/1263#issuecomment-573091956)

Issue #1282, opened 2019-09-30
SearchMessage input closes immediately after focus


pjcarly:
After upgrading to version version 3.0.4, I noticed a behaviour change in the application that I cant immediately figure out.

I have a component like this:
```
<PowerSelect 
  @class="search-input" 
  @searchMessage="Search"  
  @searchMessageComponent={{component "recently-viewed" recordSelected=(action "recentlyViewedRecordSelected")}}
  @placeholder={{@placeholder}} 
  @search={{perform this.searchTask}} 
  @searchEnabled={{true}}
  @onChange={{action "valueChanged"}} 
  @onFocus={{action "handleFocus"}} 
  @renderInPlace={{true}} as |model|>
  <div class="search-result">
    <div class="name">
      {{model.name}}
    </div>
    <div class="type">
      {{model.constructor.modelName}}
    </div>
  </div>
</PowerSelect>
```

On focus of the Component, I want to display a list of recently viewed records, which I could accomplish with `@searchMessageComponent` (this does seem to be missing from the documentation, but is still in the codebase)

The issue i'm seeing now, is that:
 * on focus, the component im trying to use is displayed and immediately closed. 
 * clicking power-select, and holding the mouse button down, keeps the component open, until release of the mouse button
 * when manually opening the search input, through the public api (`element.actions.open();`) it works as it should.

I narrowed down the code that closes the addon to here:
https://github.com/cibernox/ember-basic-dropdown/blob/b60127457821069cf288a7008c6ec2f787a69497/addon/components/basic-dropdown-content.js#L125

pjcarly:
After some more tinkering, I managed to relate it to the `@renderInPlace={{true}}` attribute, this only happens with renderInPlace set to true

pjcarly:
Alright. Narrowed it down completely.

If you apply the CSS styling
`top: 0` on the `ember-basic-dropdown-content` element (in order for the input element to cover the trigger completely). The issue happens.

My guess is that chrome somehow triggers a click element on the document, and incorrectly assumes that the click is outside the select element.

cibernox:
@pjcarly do you have any idea of what the fix would be? I also fail to see why this would have changed in 3.0. The logic is mostly a 1:1 translation to new patterns, but the same events are being handled.

pjcarly:
I have no idea, only thing I can say is that I upgraded from v2.2.2, and there it is working with `top: 0`. For now I have disabled the `top: 0 and I can use it.

[...]

Issue #1263, opened 2019-08-14
Menu closes prematurely when clicking trigger bottom edge


mayatron:
This is a minor thing, but the select menu appears to open and immediately close (with a single click) when clicking the very bottom edge of the trigger. See example below.

![ezgif-1-6359e74e004b](https://user-images.githubusercontent.com/1009725/62986775-f5326780-bdf1-11e9-9c91-01b2b81d024d.gif)

mbernwieser:
I could observe the same behavior, but not only when you click on the edge, but especially when you click on the select box after losing focus.

This only occurs since version 3.0.0. If I use version 2.3.5 this does not occur.

![Screen Capture](https://user-images.githubusercontent.com/18480335/64962962-ad8f7780-d898-11e9-88ab-b994106f0b48.gif)

cibernox:
@mbernwieser Interesting regression. Are you able to reproduce it in any of the examples in the docs? I ask because it would be easier to track with a reproduction

mbernwieser:
@cibernox unfortunately not (or perhaps better so ;) ). I'll see if I find time to track the bug down in the next few days

bitwolfe:
I ran into this issue as well. By my testing it seems to be related to the eventType changes to BasicDropdown or something. The search field being enabled also seems to prevent the issue from showing itself, and so naturally the multiple select is in worse shape due to it not having a search field at the top of the dropdown.

Haven't had the time to debug if events are fired multiple times or not. Also unfortunately unable to reproduce this in the docs.

Tested in Microsoft Edge Beta (Chromium).

YvesBaldus:
@cibernox I think this happens because the rootEventType of the basic dropdown is "click" and the eventType for the trigger is "mousedown". If you click the very bottom edge of the trigger, the trigger will get the mousedown event and the content opens. Because we are on the very edge of the trigger and the content is open the content will actualy receive the mouseup event. 

Therefore the click event fires on the most specific ancestor of both elements. In our case this would propably the body element, if the content is wormholed. The dropdown registers a click event outside of itself and closes again.

![dd-bug](https://user-images.githubusercontent.com/10119177/70133710-f16b6f00-1686-11ea-861a-05c2f0b682df.gif)

cibernox:
@ybaldus interesting. In the latest beta I changed the default to be click, can you install 4.0.0-beta and check if the problem is fixed for you? It requires Ember 3.13 tho.

ybaldus:
[...]

PR #1335, opened 2020-02-26
Fix premature close with content placed over trigger

mayatron:
### Related issues

- Addresses https://github.com/cibernox/ember-power-select/issues/1334
- Addresses https://github.com/cibernox/ember-power-select/issues/1282
- Addresses https://github.com/cibernox/ember-power-select/issues/1263

### Background

Using a value of `mousedown` for the `eventType` that is set on the Ember Basic Dropdown trigger is problematic in some cases, especially when content overlaps the trigger. The reason for this is that the subsequent `mouseup` event handler that is set in the `addHandlers` action in the `options.ts` file is triggered immediately after the `mousedown` that opens the dropdown content in the first place. The `mouseup` event handler calls `findOptionAndPerform` which will select any valid, (though in this case accidentally) highlighted option and then immediate close the dropdown when it makes the selection.

Note that even if the above `mouseup` event listener is bypassed, Ember Basic Dropdown _will still immediately close_ the dropdown, though this time without Power Select having made a selection. This happens because, in the `basic-dropdown-content.ts` file, in the `setup` action adds a `click` event listener (based on the value of `this.args.rootEventType`) to the document, to ensure the content is closed when the user clicks away from the menu.... but when the resulting `handleRootMouseDown` function is called the `click` event target is `body`, which signals EBD to close the content. It's not clear to me why the `target` is `body` in this function, but I suspect it may be related to the `mousedown` and `mouseup` targets differing here, when the content appears above the trigger.

Specifying `@eventType='click'` instead bypasses both problems because `click` is fired _only after_ the `mousedown` and `mouseup` events, resulting in more predictable behavior, regardless of location of the content in relation to the trigger.

### Other Notes

- Also fixes unrelated, intermittent test failures

### Related, alternative PR

- https://github.com/cibernox/ember-power-select/pull/1333

### DIFF
--- a/addon/templates/components/power-select.hbs
+++ b/addon/templates/components/power-select.hbs
@@ -24,7 +24,7 @@
     actions=(assign dropdown.actions this._publicAPIActions)
   )) (concat "ember-power-select-options-" dropdown.uniqueId) as |publicAPI listboxId|}}
     <dropdown.Trigger
-      @eventType={{or @eventType "mousedown"}}
+      @eventType={{or @eventType "click"}}
       {{did-insert this._updateOptions @options}}
       {{did-update this._updateOptions @options}}
       {{did-insert this._updateSelected @selected}}
--- a/tests/dummy/app/templates/public-pages/docs/api-reference.hbs
+++ b/tests/dummy/app/templates/public-pages/docs/api-reference.hbs
@@ -146,7 +146,7 @@
       <td>eventType</td>
       <td><code>string</code></td>
       <td>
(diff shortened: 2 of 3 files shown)


mayatron:
@cibernox let me know if you have any feedback or thoughts on this. Also happy to split out the test fixes into a separate PR if you prefer that.

mayatron:
Also note there are some unrelated test failures for some ember-try build scenarios. These started occurring recently in other PRs as well, for example: https://travis-ci.org/cibernox/ember-power-select/jobs/653542073?utm_medium=notification&utm_source=github_status

mayatron:
Closing my PR due to lack of activity.

PR #1337, opened 2020-03-04
wrap spellcheck boolean in curlies

uconnjosh:
together with: @krainboltgreene

ensures that `spellcheck` is set to "false"


cibernox:
Sorry for the delay, parenthood and life in general ate all my attention lately. This is a great catch!

I'll release a new version soon, hopefully today.

Issue #1341, opened 2020-03-23
v4-beta ember-power-select-is-selected helper not working properly with ember-data collections


tobiaszwalczak:
Since upgrading to EPS `4.0.0-beta.6` (to resolve the issue https://github.com/cibernox/ember-power-select/issues/1338), I experience problems with the `ember-power-select-is-selected` helper when passing an ember-data collection/relation to the `@selected` attribute.

I investigated the issue and in this version, the `ember-power-select-is-selected` helper receives the `@selected` attribute not as plain array anymore, like in the previous version `3.0.6`.

To resolve the problem I would transform the `@selected` attribute with `toPlainArray` also when it is assigned from the arguments:

https://github.com/cibernox/ember-power-select/blob/master/addon/components/power-select.ts#L200

Like it is already handled for `@options`:

https://github.com/cibernox/ember-power-select/blob/master/addon/components/power-select.ts#L186

cibernox:
This can be closed now

PR #1342, opened 2020-03-23
Apply toPlainArray on @selected for publicAPI

tobiaszwalczak:
Fix for: https://github.com/cibernox/ember-power-select/issues/1341

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 # Master
+- [BUGFIX] When `@selected` is an ember-data collection, transform to a plain array for the public API
 
 # 4.0.0-beta.6
 - [BUGFIX] Multiple selects can receive an ember-data collection in the `@selected` attribute.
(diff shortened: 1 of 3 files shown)


cibernox:
@tobiaszwalczak good catch. However I don't get why this change would require to change an existing test, why is that?

tobiaszwalczak:
@cibernox Ah right, that goes maybe a bit far for this PR. As I fixed the `@selected` attribute to resolve an issue (https://github.com/cibernox/ember-power-select/issues/1341) with the `ember-power-select-is-selected` helper and the `aria-selected` attribute it influences, I wanted the test to reflect that the `aria-selected` attribute works properly. The test should still reflect the previous functionality it tested.

But I'm happy to revert the test file if you want me to 😄

cibernox:
I believe I've just fixed this in master.

cibernox:
I just released 4.0.0 final including this change, can you give it a go?

tobiaszwalczak:
@cibernox Just tried out `4.0.0`and works like a charm now, thanks! 👌 Going to close the PR and the related Issue.

Issue #1352, opened 2020-04-16
AfterOptionsComponent is missing PublicAPI


huberts:
In commit 593ea1365cb560d3e619f94ee875f093a2677c67 the `select `parameter was cut off in `AfterOptionsComponent`:
[ember-power-select.hbs](https://github.com/cibernox/ember-power-select/blob/master/addon/templates/components/power-select.hbs#L118)
I assume it is a bug?

huberts:
I do not know the drill - should I close the issue? :)

cibernox:
Yeah, this can be closed. Thanks!

PR #1353, opened 2020-04-16
#1352 Expose PublicAPI to AfterOptionsComponent

huberts:
It fixes the issue #1352 (missing `@select` after code refactoring).


cibernox:
It totally is a bug! Thanks!

PR #1358, opened 2020-04-23
move option template into power-select/option component

gabrielgrant:
This serves as the basis for a PR to provide an option for groups to be selected.

Submitting this separately, since it's hopefully an un-controversial refactoring, whereas the other has some non-trivial API design questions

related -- moving groupName to a component: #467

### DIFF
--- a/addon/templates/components/power-select.hbs
+++ b/addon/templates/components/power-select.hbs
@@ -106,6 +106,7 @@
             @optionsComponent={{Options}}
             @extra={{@extra}}
             @highlightOnHover={{this.highlightOnHover}}
+            @optionComponent={{or @optionComponent "power-select/option"}}
             @groupComponent={{or @groupComponent "power-select/power-select-group"}}
             id={{listboxId}}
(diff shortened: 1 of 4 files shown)


cibernox:
It is a bit mor controversial that one may think. The reason for the option not being a component by default was really all about speed. Selects with a handful of elements won't notice it, but selects with a few hundreds may take longer to render. I'll checkout this branch and check if it does affect performance or being a template-only component makes it fast enough to not be an issue anymore.

cibernox:
Sorry for the delay, but I finally had time to test it. Rendering 1000 elements (which I agree is something nobody should be doing) opening the component goes from 2s to 3.7s.

I feel the slowdown is too big for this to be a default. I'd suggest that if someone needs to customize how each individual options is rendered they can provide a parent `@optionsComponent`, which includes the `{{#each}}` loop user for each element.

A bit more inconvenient, but in this case I err on the side of performance.

Thank your for your PR, and I hope performance is good enough that we don't have to worry about it.

PR #1363, opened 2020-05-03
Fixes #1362: Error thrown when promise of selected argument changes

### LINKED ISSUE #1362: Error thrown when PromiseProxy on "selected" argument is changed
When changing the PromiseProxy on the select argument of power select, Ember throws the following error:
```
Error: Assertion Failed: You attempted to remove a function listener which did not exist on the instance, which means you may have attempted to remove it before it was added.
```

How the reproduce:
```hbs
{{!-- app/templates/visit.hbs --}}
<PowerSelect @selected={{this.model.cities}} @onChange={{fn (mut this.model.cities)}} as |name|>
  {{name}}
</PowerSelect>
```

```javascript
// app/routes/visit.js
import Route from '@ember/routing/route';

export default class VisitRoute extends Route {
    model(params) {
        return this.store.findRecord('visit', params.visit_id);
    }
}


// app/models/visited.js
import Model from '@ember-data/model';

export default class VisitModel extends Model {
	@hasMany('city') cities
}


// app/models/city.js
import Model from '@ember-data/model';

export default class CityModel extends Model {
	@attr('string') name
}
```

On transition from Visit id=1 to Visit id=2, Ember will throw given error.

adezegher:
Fixed missing ``target`` argument in calling removeObserver

### DIFF
--- a/addon/components/power-select.ts
+++ b/addon/components/power-select.ts
@@ -335,7 +335,7 @@ export default class PowerSelect extends Component<PowerSelectArgs> {
       let currentSelectedPromise: PromiseProxy<any> = this.args.selected;
       if (Object.hasOwnProperty.call(currentSelectedPromise, 'content')) { // seems a PromiseProxy
         if (this._lastSelectedPromise) {
-          removeObserver(this._lastSelectedPromise, 'content', this._selectedObserverCallback);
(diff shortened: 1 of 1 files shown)


cibernox:
Ooooooh! I think this is going to fix a problem I had a few times. I looked at the code a hundred times and I missed that!

cibernox:
Thanks!

PR #1373, opened 2020-06-22
Update ember-concurrency-decorators@2.0.0

mydea:
With the major release of ember-concurrency-decorators, you will get dependency-lint errors if trying to update it, as 2.0.0 is out of range as defined in this addon.

This PR updates this to 2.0.0 (and ember-concurrency to 1.2.0 along, as that brings brings the types for ember-concurrency-decorators 2.0.0).

As a sidenote, there are both yarn.lock as well as package-lock.json files checked in, both modified at different times. I've updated the yarn.lock files now, but it would be a good idea to remove one or the other.

Finally, this also cleans up the duplicate @glimmer/component dependency (it was defined twice).

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
     "ember-cli-htmlbars": "^4.2.3",
     "ember-cli-typescript": "^3.1.1",
     "ember-concurrency": "^1.2.0",
-    "ember-concurrency-decorators": "^1.0.0",
+    "ember-concurrency-decorators": "^2.0.0",
     "ember-text-measurer": "^0.6.0",
     "ember-truth-helpers": "^2.1.0"
   },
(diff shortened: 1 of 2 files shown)


andreyfel:
@mydea did you have a chance to check why tests are failing?

cibernox:
@mydea if you can rebase this I believe you should have no conflicts and maybe CI will be green.

andreyfel:
This PR updates yarn.lock which was removed at #1376. It should switch to updating package-lock.json.
I've created another PR #1377 which does that and some more just in case.

mydea:
I guess this can be closed in favor of #1377 then?

cibernox:
> in favor of #1377 then?

If that's the case #1377 has just been merged.

mydea:
Awesome, I will close this then, thanks!

PR #1376, opened 2020-06-26
Fix tests related to aria-expanded

andreyfel:
The change in ember-basic-dropdown@3.0.3 made a change to aria-expanded attribute behavior. If dropdown is not expanded the attribute is still
added but with value false. See this PR for the context.
https://github.com/cibernox/ember-basic-dropdown/pull/556

Update min requirement to ember-basic-dropdown@3.0.3, fix tests.

Using package-lock.json and yarn.lock together doesn't make sense.
Remove yarn.lock since it is not used on CI therefore it doesn't make
much sense to keep it.


andreyfel:
@cibernox the build is green, could you please take a look?

cibernox:
Sorry for the delay, great catch!

cibernox:
Merged! 🎉

PR #1377, opened 2020-06-27
Update deps and cleanup

andreyfel:


andreyfel:
@cibernox, I've updated all dependencies except `ember-code-snippet` which is a breaking change and will require some refactoring and would also require to update `ember-code-example-component`.
Please let me know if you want me to revert something.

Netlify deploy has crashed with the error:
```
11:52:57 AM: Build Error (broccoli-persistent-filter:Babel > [Babel: @ember-data/store]) in -private/system/references/reference.ts
11:52:57 AM: /opt/build/repo/-private/system/references/reference.ts: Imported DEPRECATE_REFERENCE_INTERNAL_MODEL from @ember-data/private-build-infra/deprecations which is not a supported flag.
```
I believe it is related to the node_modules busted cache. Could you please wipe the cache and restart the deploy?

cibernox:
Merged! Thank you very much! I bursted the cache and I think next deploy will be fine!

PR #1380, opened 2020-07-03
Avoid unnecessary listbox attribute

mydea:
There is `role='listbox'` declared on some lists that don't need it. In fact axe complains about them. This is easily fixed by simply removing the role, which doesn't improve accessibility in this case anyhow - this can simply be a normal `<ul>` list.


mydea:
Hmm, this is actually not correct. I think the actual issue was that the lisboxes where missing `id` and a label (e.g. `aria-label`). I have instead added this to this PR:
https://github.com/cibernox/ember-power-select/pull/1306

PR #1384, opened 2020-07-28
When using a synchronous custom search function, the first result does not get highlighted

jeffhertzler:
This is a regression from previous versions of power-select


cibernox:
@jeffhertzler great catch! Ignore the failures in beta and canary. Merging!

PR #1388, opened 2020-08-20
Upgrade @glimmer/component to avoid deprecations with ember 3.20

kstinson14:
After upgrading my apps to ember 3.20, deprecation warnings began appearing due to the use of `setSourceDestroying`.  It seems these were stemming from dependencies of `@glimmer/component` lower than version 1.0.1 (the deprecation warnings were addressed in that [version](https://github.com/glimmerjs/glimmer.js/releases/tag/v1.0.1)).

Upgrading the `@glimmer/component` version here, as well as `ember-basic-dropdown` (who updated `@glimmer/component` [recently](https://github.com/cibernox/ember-basic-dropdown/pull/571)) should remove these deprecations in apps using ember 3.20.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9011,9 +9011,9 @@
       }
     },
     "ember-basic-dropdown": {
-      "version": "3.0.9",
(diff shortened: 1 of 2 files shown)


cibernox (approved):

cibernox:
Good catch, thanks!

PR #1389, opened 2020-08-31
fix wrong component name in cookbook

luxzeitlos:
The cookbook says you should create a file `ember-power-select.js` for global configuration, but the file must be named `power-select.js`.


mkszepp:
Unfortunately I can no longer merge this it will be fixed with #1389

PR #1390, opened 2020-09-03
colocate component templates

jelhan:
Ember Power Select was relying on implicit convetions for template lookups. This is working fine unless a consumer tries to rename the component. In that case the template lookup breaks. This could be avoided by explicitly binding the component to it's template.

The Octane way to achive this is colocating the component templates. This will bound the template to the component at build-time and prevent any run-time lookup.

This enables a consumer to rename the component to avoid naming collisions. E.g. a customized version of Ember Power Select can be used with the `<PowerSelect>` name. This would look like this:

```js
// app/components/power-select/original.js

export { default } from 'ember-power-select/components/power-select';
```

```hbs
// app/components/power-select.hbs

<PowerSelect::Original
  @options={{@options}}
  @selected={{@selected}}
  @onChange={{@onChange}}
/>
```

It also makes maintaining the addon easier. But I was more interested in the consumer perspective to be honest.

I don't think this is breaking change. Ember Power Select already requires Ember >= 3.13. Component template colocation landed in Ember 3.13.

The readme does not explicitly mention Ember CLI compatibility. But I guess that means that this is included in the general statement about Ember compatibility?

To be honest I'm not sure what minimum Ember CLI version is required for component template colocation. But I'm sure that Ember CLI 3.13+ will work.


cibernox:
It's true that ember-cli compatibility was not explicitly mentioned, but that at least requiring the same version of Ember-cli as the one required of ember itself (3.13) was sort of implied. And since the most recent version is 3.21, it's also very reasonable (3.13 has almost two human years, witch is roughy 60 javascript years).

Merging!

PR #1392, opened 2020-09-04
[bugfix] remove remaining observer on PowerSelect teardown

### LINKED ISSUE #1391: JS Heap grows with selected PromiseProxy
Hello and thanks for the great addon.

We are exercising this [block of code that observes PromiseProxies](https://github.com/cibernox/ember-power-select/blob/fb688e44c166e42e58d6c89de02ede553464e38a/addon/components/power-select.ts#L335-L341) and noticed that `PowerSelect` Component instance accumulates in the JS Heap when the initial `selected` value is a `PromiseProxy`. 

I believe that in Chrome, these Component instances are not torn down until a Major Garbage Collection occurs in the browser (Chrome calls this "Major GC" in their Performance dev tools flame chart).

**Dummy App Reproduction Steps**

I've modified the "Playground" portion of the app to show the Component accumulation when the initial `selected` value is a PromiseProxy. The "Playground" renders one PowerSelect component.

```hbs
{{!-- dummy/app/templates/playground.hbs--}}
<h1>Playground</h1>
<PowerSelect @selected={{this.selected}} @onChange={{fn (mut this.selected)}} as |name|>{{name}}</PowerSelect>
```

```js
// dummy/app/controllers/playground.js
import ObjectProxy from '@ember/object/proxy';
import PromiseProxyMixin from '@ember/object/promise-proxy-mixin';
let ObjectPromiseProxy = ObjectProxy.extend(PromiseProxyMixin);

export default class extends Controller {
  init() {
    super.init(...arguments);

    const proxy = ObjectPromiseProxy.create({ promise: resolve('I am a PromiseProxy Object') });
    this.set('selected', proxy);
  }
}
```

The initial state of the JS Heap contains**1** PowerSelect instance, but the final state contains **6** instances (because we visited a page containing the PowerSelect with `selected` PromiseProxy a total of five times). Although the component is destroyed, it still remains in the JS Heap. 

_Visiting a page containing  a `PowerSelect` where `selected` value is a PromiseProxy (visited 5 times in the gif)_
Screen Recording 2020-09-04 at 09 58 AM

flexyford:
PowerSelect Component instances accumulate in the JS Heap when the initial `selected` value is a PromiseProxy. That's because we [addObserver for PromiseProxies](https://github.com/cibernox/ember-power-select/blob/fb688e44c166e42e58d6c89de02ede553464e38a/addon/components/power-select.ts#L335-L341), but the observer is not removed prior to component teardown.

We now remove the observer on component teardown, which allows PowerSelect Component instances to be removed from the JS Heap when the Component is destroyed.

Fixes #1391

---

The initial state of the JS Heap contains **1** PowerSelect instance, and the final state contains **1** instances (even though we visited a page containing the PowerSelect with `selected` PromiseProxy a total of five times). When the component is destroyed, it is removed from the JS Heap. 

Screen Recording 2020-09-04 at 11 25 AM

### DIFF
--- a/addon/components/power-select.ts
+++ b/addon/components/power-select.ts
@@ -349,9 +349,13 @@ export default class PowerSelect extends Component<PowerSelectArgs> {
       }
 
       let currentSelectedPromise: PromiseProxy<any> = this.args.selected;
-      if (isPromiseProxyLike(currentSelectedPromise)) {
-        addObserver(currentSelectedPromise, 'content', this, this._selectedObserverCallback);
-      }
+      currentSelectedPromise.then(() => {
+        if (this.isDestroyed || this.isDestroying) return;
+        if (isPromiseProxyLike(currentSelectedPromise)) {
+          addObserver(currentSelectedPromise, 'content', this, this._selectedObserverCallback);
+        }
+      });
+
       this._lastSelectedPromise = currentSelectedPromise;
       this._lastSelectedPromise.then(resolvedSelected => {
         if (this._lastSelectedPromise === currentSelectedPromise) {
--- a/addon/components/power-select.ts
+++ b/addon/components/power-select.ts
@@ -336,11 +336,12 @@ export default class PowerSelect extends Component<PowerSelectArgs> {
     if (!this.args.selected) return;
     if (typeof this.args.selected.then === 'function') {
       if (this._lastSelectedPromise === this.args.selected) return; // promise is still the same
+      if (this._lastSelectedPromise && isPromiseProxyLike(this._lastSelectedPromise)) {
+        removeObserver(this._lastSelectedPromise, 'content', this, this._selectedObserverCallback);
+      }
+
       let currentSelectedPromise: PromiseProxy<any> = this.args.selected;
       if (isPromiseProxyLike(currentSelectedPromise)) {
-        if (this._lastSelectedPromise) {
-          removeObserver(this._lastSelectedPromise, 'content', this, this._selectedObserverCallback);
-        }
         addObserver(currentSelectedPromise, 'content', this, this._selectedObserverCallback);
       }
       this._lastSelectedPromise = currentSelectedPromise;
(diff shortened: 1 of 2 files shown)


flexyford:
Closing while I look into failing tests

cibernox (approved):

cibernox:
One of the errors in CI seems a legit edge case. Can you take a look?

flexyford:
@cibernox I learned that we should only add an observer to the `content` property of a PromiseProxy when the underlying Promise object is fulfilled. Otherwise, the `content` property does not exist and we're not actually observing any changes to the `content` property. 

```js
const deferred = RSVP.defer();
const promiseProxy = ObjectPromiseProxy.create({ promise: deferred.promise });

promiseProxy.isFulfilled; // => false
promiseProxy.hasOwnProperty('content') // => false

deferred.resolve();

promiseProxy.isFulfilled; // => true
promiseProxy.hasOwnProperty('content') // => true
```

The edge-case test failure was attempting to remove an observer on `willDestroy()` because the PromiseProxy was fulfilled at the point, but that observer was never added because the PromiseProxy was pending when we first entered `_updateSelected()`.

cibernox:
Great, thanks! You can safely ignore the bugs in beta and canary, they are unrelated.

cibernox:
@flexyford you mentioned one more memory leak, should I expect another PR for that? if so, I will wait a bit before publishing a new version to NPM.

flexyford:
> @flexyford you mentioned one more memory leak, should I expect another PR for that? if so, I will wait a bit before publishing a new version to NPM.

@cibernox yes, I plan to open another PR for that this week. Does that sound like an appropriate timeline to wait for a release?

cibernox:
In that case I'll release a new version now and I'll release another one whenever it's done. I'd have waited if you happened to have it almost ready, but versions are cheap anyway.

cibernox:
released 4.0.4 🎉

PR #1394, opened 2020-09-09
[bugfix] Remove event listeners on destroy

### LINKED ISSUE #1393: Js Heap contains extra PowerSelect instance when default Options Component rendered
We are exercising this `PowerSelect/Options` Component and noticed that the [eventListeners of PowerSelect/Options](https://github.com/cibernox/ember-power-select/blob/fb688e44c166e42e58d6c89de02ede553464e38a/addon/components/power-select/options.ts#L57-L72) are not never removed. This causes an extra instance of the `PowerSelect` Component to remain in the JS Heap until the `PowerSelect/Options` Component is re-rendered. 

**Dummy App Reproduction Steps**

I've modified the "Playground" portion of the app to show the Component accumulation when the Options dropdown is rendered. The "Playground" renders one PowerSelect component.

```hbs
{{!-- dummy/app/templates/playground.hbs--}}
<h1>Playground</h1>
<PowerSelect @options={{this.cities}} @selected={{this.selected}} @onChange={{fn (mut this.selected)}} as |name|>{{name}}</PowerSelect>
```

The initial state of the JS Heap contains **1** PowerSelect instance, but the final state contains **2** instances. Although the component is destroyed, it still remains in the JS Heap. This is not the same linear component instance JS Heap growth we saw in #1391 but is a memory leak nonetheless that we can patch up. The remaining instance is removed when the Options dropdown is re-rendered. 

_Visiting a page containing a `PowerSelect`_ 
    **Snapshots explained:**
  - Snapshot 1: Visit PowerSelect Playground and open `Options` dropdown
    - **1** PowerSelect instance exists
  - Snapshot 2: Visit Addons Page
    - **2** PowerSelect instances exist (instance from Snapshot 1 remains)
  - Snapshot 3: Visit PowerSelect Playground
    - **2** PowerSelect instances exist (instance from Snapshot 1 remains)
  - Snapshot 4: Open `Options` dropdown (re-rendered)
    - **1** PowerSelect instances exist (instance from Snapshot 1 removed)

  Screen Recording 2020-09-08 at 11 28 AM

flexyford:
An extra `PowerSelect` Component instance remains in the JS Heap until the `PowerSelect/Options` Component is re-rendered. That's because we [addEventListeners in PowerSelect/Options](https://github.com/cibernox/ember-power-select/blob/fb688e44c166e42e58d6c89de02ede553464e38a/addon/components/power-select/options.ts#L57-L72) but the listeners are not removed prior to component teardown.

This PR eagerly removes all event listeners on component teardown, which allows PowerSelect Component instances to be removed from the JS Heap when the Component is destroyed.  This implementation may attempt to remove listeners that have not been added. This is okay since calling removeEventListener() with arguments that do not identify any currently registered EventListener on the EventTarget has no effect.

Fixes #1393 

---

The initial state of the JS Heap contains **1** PowerSelect instance, and only **1** instance is ever allocated at any one time in memory throughout the life of the demonstration. When the component is destroyed, it is removed from the JS Heap.

Screen Recording 2020-09-08 at 08 13 PM

### DIFF
--- a/addon/components/power-select/options.ts
+++ b/addon/components/power-select/options.ts
@@ -34,6 +34,11 @@ if(typeof FastBoot === 'undefined'){
 export default class Options extends Component<Args> {
   private isTouchDevice = isTouchDevice
   private hasMoved = false
+  private mouseOverHandler = (_: MouseEvent): void => {}
+  private mouseUpHandler = (_: MouseEvent): void => {}
+  private touchEndHandler = (_: TouchEvent): void => {}
+  private touchMoveHandler = (_: TouchEvent): void => {}
+  private touchStartHandler = (_: TouchEvent): void => {}
 
   @action
   addHandlers(element: Element) {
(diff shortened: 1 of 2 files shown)


addon/components/power-select/options.ts
```diff
@@ -34,6 +34,11 @@ if(typeof FastBoot === 'undefined'){
 export default class Options extends Component<Args> {
   private isTouchDevice = isTouchDevice
   private hasMoved = false
+  private mouseOverHandler = (_: MouseEvent): void => {}
+  private mouseUpHandler = (_: MouseEvent): void => {}
+  private touchEndHandler = (_: TouchEvent): void => {}
+  private touchMoveHandler = (_: TouchEvent): void => {}
+  private touchStartHandler = (_: TouchEvent): void => {}
```
flexyford, line 41: This declaration seemed to make satisfy the TS compiler. Is this an appropriate way to declare a Function in Typescript that is meant to be overwritten at runtime.

flexyford:
@cibernox anything I can do here to help get this to the finish line?

addon/components/power-select/options.ts
```diff
@@ -54,22 +59,24 @@ export default class Options extends Component<Args> {
...
+      this.mouseOverHandler = (e: MouseEvent): void => findOptionAndPerform(this.args.select.actions.highlight, this.args.select, e);
+      element.addEventListener('mouseover', this.mouseOverHandler);
     }
     if (this.isTouchDevice) {
-      let touchMoveHandler = () => {
+      this.touchMoveHandler = (_: TouchEvent): void => {
         this.hasMoved = true;
         if (element) {
-          element.removeEventListener('touchmove', touchMoveHandler);
+          element.removeEventListener('touchmove', this.touchMoveHandler);
```
cibernox, line 72: Just one suggestion. Can you declare these event handlers that are not used in templates as _private_?

cibernox (approved):

addon/components/power-select/options.ts
cibernox: Looks correct to me.

cibernox:
LGTM, I just added a small suggestin. Sorry for not merging this before, it just slept under my radar and I forgot about this completely.

addon/components/power-select/options.ts
cibernox: Oh, right. Nevermind then.

cibernox:
thanks! I'll release a new patch version later today

cibernox:
@flexyford released 4.0.5! 🎉

PR #1399, opened 2020-09-24
Support both ember-concurrency 1.x and 2.x

maxfierke:
We've just released ember-concurrency 2.0.0-beta.1. It's a largely API-compatible re-engineering of ember-concurrency's internals to support tracked properties and decouple from Ember (shedding EmberObject, etc.). As a popular library which depends on ember-concurrency, it would be nice for `ember-power-select` to support both 1.x and 2.x versions concurrently so that folks can upgrade the version of ember-concurrency in their applications at their leisure without waiting for all the add-ons they use to switch over to a newer version (and risk leaving folks who cannot upgrade yet behind.)

This PR adds some scenarios to ember-try to enable testing against both major releases and relaxes the version restriction.

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
     "ember-cli-babel": "^7.21.0",
     "ember-cli-htmlbars": "^5.2.0",
     "ember-cli-typescript": "^3.1.4",
-    "ember-concurrency": "^1.2.1",
+    "ember-concurrency": ">=1.0.0 <3",
     "ember-concurrency-decorators": "^2.0.0",
     "ember-text-measurer": "^0.6.0",
     "ember-truth-helpers": "^2.1.0"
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -55,7 +55,7 @@
     "ember-cli-babel": "^7.21.0",
     "ember-cli-htmlbars": "^5.2.0",
     "ember-cli-typescript": "^3.1.4",
-    "ember-concurrency": "^1.2.1",
+    "ember-concurrency": "^1.0.0 <3",
```
Turbo87, line 58: according to the [Online Semver checker](https://jubianchi.github.io/semver-check/#/^1.0.0%20<3/2.0.0) this does not quite work, because `2.0.0` is not matching that constraint

maxfierke: oh whoops, yep. that should be `>=1.0.0 <3`

cibernox: I believe you can do `"^1.0.0 | ^2.0.0"`

maxfierke: @cibernox I just updated it to use `>=1.0.0 <3` but let me know if you have a preference for `"^1.0.0 || ^2.0.0-beta.1"`

cibernox: Not really, I believe both work.

cibernox:
This looks good, the failures in beta and canary are not because of your changes, I need to track those down tomorrow.

Thanks!

PR #1406, opened 2020-10-16
Add support for stylus (Apple Pencil)

### LINKED ISSUE #1387: Apple Pencil tap does not work
First of all, thanks a lot for such a useful addon!

I'm reporting it here, but it's relevant for `ember-basic-dropdown` as well.
Here is the GIF illustrating the problem:

apple_pencil_issue

A single tap using a finger triggers two events: `touchstart` and `touchend`.
A single tap using an apple pencil also triggers `touchmove` event because of pencil being very precise — it's just a few pixels of touch "distance", but current logic aborts an action if `touchmove` event was triggered at all.

Here is the list of all code occurrences that I've found:
- [ember-power-select/addon/components/power-select/options](https://github.com/cibernox/ember-power-select/blob/c36d826efd9bef2c732f6d1b6733af5bc76cef00/addon/components/power-select/options.ts#L77-L80)
- [ember-basic-dropdown/addon/components/basic-dropdown-trigger](https://github.com/cibernox/ember-basic-dropdown/blob/ad81ef4dc161943bdc05ac3048b209f30ee19a4b/addon/components/basic-dropdown-trigger.ts#L101-L103)
- [ember-basic-dropdown/addon/components/basic-dropdown-content](https://github.com/cibernox/ember-basic-dropdown/blob/ad81ef4dc161943bdc05ac3048b209f30ee19a4b/addon/components/basic-dropdown-content.ts#L107-L110)

krasnoukhov:
Fixes #1387 

We've been using a similar patch in our apps for a while and have received no negative feedback 😅


addon/components/power-select/options.ts
```diff
@@ -32,8 +32,8 @@ if(typeof FastBoot === 'undefined'){
 }
 
 export default class Options extends Component<Args> {
-  private isTouchDevice = isTouchDevice
-  private hasMoved = false
+  private isTouchDevice = (this.args as any)?.extra?._isTouchDevice || isTouchDevice
```
cibernox, line 35: do you need this? In the `testem.js` file we seem to start the browser with the `--touch-events` flag, so that should be enough, right?

cibernox (changes requested):

addon/components/power-select/options.ts
```diff
@@ -119,4 +119,19 @@ export default class Options extends Component<Args> {
     }
     return option;
   }
+
```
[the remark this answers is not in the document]
krasnoukhov: Thanks for the quick feedback! Yes, good idea. My testing showed that using `5` works well

addon/components/power-select/options.ts
krasnoukhov: Not sure what other approach can be taken here. In [ember-basic-dropdown](https://github.com/cibernox/ember-basic-dropdown/pull/582/files#diff-47a26d7d1178b224b35da72ba32c96aea5490b6bfa7662fbd52286e1ed39a083R286) this option is a part of component API... Not sure if that's better honestly 😅 

But yeah, this is only needed for testing.

cibernox (approved):

krasnoukhov:
Hi @cibernox, sorry to ping you, but do you want to merge this since you approved? I merged upstream changes to the build is green.

Please note that this should go together with ember-basic-dropdown PR: https://github.com/cibernox/ember-basic-dropdown/pull/582

rwwagner90:
@krasnoukhov I know this has been open awhile. Any chance you could rebase and resolve the conflicts please?

krasnoukhov:
@rwwagner90 Done. Tests are passing locally

addon/components/power-select/options.ts
rwwagner90, line 35: @krasnoukhov shouldn't the args not be `any` here? We have types for the args.

rwwagner90 (approved):

PR #1410, opened 2020-10-22
Widen `ember-truth-helpers` dependency constraint to allow v3.x

Turbo87:
same as https://github.com/cibernox/ember-basic-dropdown/pull/584

This avoids users having to use resolutions to update to ember-truth-helpers v3 :)


Turbo87:
looks like the CI failure is unrelated to this change and caused by changes in the current Ember version

cibernox:
Thanks for all the PRs!

PR #1411, opened 2020-10-22
package.json: Move `@ember-decorators/component` into `devDependencies`

Turbo87:
Now that we use glimmer components this dependency is no longer needed. There is still one usage in the dummy app though, so we can move it to `devDependencies` for now.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,6 @@
     "vendor"
   ],
   "dependencies": {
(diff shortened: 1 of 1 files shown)


Turbo87:
looks like the CI failure is unrelated to this change and caused by changes in the current Ember version

PR #1413, opened 2020-10-22
onChange action receives Event as last arg on mouseup

### LINKED ISSUE #1412: Unexpected argument passed-to onChange action
Hello Power-Select maintainers 👋 

We're seeing an unexpected argument passed to the `onChange` action. 

_Background_

We're using power-select-multiple and want to allow our users to `Shift + Click` to bulk select multiple items in a list, much like a [native `<select>` allows you to do](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select#Selecting_multiple_options). 

To achieve this, we're relying on the `Event` passed to `onChange`. When selecting an option in the dropdown, the `mouseup` event is fired, and we call the `onChange` action. 

_Bug_

The `onChange` action has the following method signature:

> onChange: (selection: any, select: Select, event?: Event) => void

 However, the third argument is not an `Event`, it's a duplicate `Select` object. 

I'll add a PR shortly which adds the `Event` object as the third argument as expected by the `onChange` method signature.

flexyford:
This change ensures that when an option in the `PowerSelect/Options` component is clicked, the action registered on `mouseup` will pass the`mouseup` `Event` all the way to the `onChoose` action.

Previously, the second AND third event passed to the onChoose action was the dropdown, aka `this.storedAPI`, due to arguments not being passed as expected from the `mouseup` handler, and the Event was never received by the `onChoose` action.

A test has been included in this change. 

Fixes #1412

### DIFF
--- a/addon/components/power-select/options.ts
+++ b/addon/components/power-select/options.ts
@@ -46,7 +46,7 @@ export default class Options extends Component<Args> {
     if (role === 'group') {
       return;
     }
-    let findOptionAndPerform = (action: Function, select: Select, e: Event): void => {
+    let findOptionAndPerform = (action: Function, e: Event): void => {
       if (e.target === null) return;
       let optionItem = (e.target as Element).closest('[data-option-index]');
       if (!optionItem) {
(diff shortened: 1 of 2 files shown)


mayatron:
I believe this may be happening with other functions as well, for example `onClose`. The expected arguments for that function are `(api, event)` but I'm seeing `(api, api)` instead.

AndreJoaquim:
@cibernox I'm seeing this as well when upgrading from `2.x` to `4.x`. I can't really tell when was this introduced, but could we get this merged? 🙏

AndreJoaquim:
Hello @cibernox! Is there any chance to get this merged? I was trying to avoid making a fork to be able to migrate to `4.x` 😢  and this is currently blocking my upgrade. Sorry for the bump

cibernox:
I checked CI and the failures are unrelated. I'm merging this.

Sorry for neglecting it for so long!

AndreJoaquim:
Thank you very much @cibernox! 😊

PR #1427, opened 2021-01-27
🐞 Fix error when providing `@initiallyOpened={{true}}` and `@selected`

### LINKED ISSUE #1426: Cannot read property 'uniqueId' of undefined when @initiallyOpened={{true}} and @selected is provided
Hey @cibernox.

I'm upgrading our Ember app from v3.17.0 to v3.20.2 and a part of that is upgrading EPS from `^v3.0.4` to `^4.1.0`.

I am finding that when I set `@initiallyOpened={{true}}` and have provided `@selected` then I get the following error:

```
Uncaught (in promise) TypeError: Cannot read property 'uniqueId' of undefined
    at PowerSelect._scrollTo (power-select.js:615)
    at Options.addHandlers (options.js:181)
    at Object.installModifier (did-insert.js:62)
```

I've knocked together a basic reproduction of this in codesandbox here -> https://codesandbox.io/s/goofy-frost-4zphd?file=/app/components/cheese.hbs

I've not managed to identify the issue as yet. I'll dig around but if you happen to know what the problem might be, that'd be ace.

Thanks man. Let me know if there is anything I can do to help.

achambers:
#1288 Seems to have introduced a bug whereby providing `@initiallyOpened` AND `@selected` causes an error as described in #1426 .

This PR adds a failing test for the bug and also the bug fix.

Closes #1426

### DIFF
--- a/addon/components/power-select.hbs
+++ b/addon/components/power-select.hbs
@@ -29,6 +29,7 @@
       {{did-update this._updateOptions @options}}
       {{did-insert this._updateSelected @selected}}
       {{did-update this._updateSelected @selected}}
+      {{did-insert this._registerAPI publicAPI}}
       {{did-update this._registerAPI publicAPI}}
       {{did-update this._performSearch this.searchText}}
       {{on "keydown" this.handleTriggerKeydown}}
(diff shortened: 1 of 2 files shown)


cibernox:
Looks good to me. This was probably an oversight during that big refactor.

PR #1430, opened 2021-02-15
Fix memory leaks / Add ember-cli-memory-leak-detector

steveszc:
This PR installs an add-on I authored called [ember-cli-memory-leak-detector](https://github.com/steveszc/ember-cli-memory-leak-detector). 
 
The add-on works by recording a heap snapshot after the suite tests finishes running. The heap snapshot is then searched for any retained instances of the host app/addon's ES classes (indicating a memory leak). If it finds retained classes in the heap snapshot, the test suite fails and a report of the retained classes is logged in the console output. 

After installing ember-cli-memory-leak-detector (and fixing the few remaining leaky codepaths) we can ensure that no memory leak is ever introduced into EPS in the future, which is a big win for a such a popular addon! 

I'll be giving a talk at EmberConf this year about ember-cli-memory-leak-detector and I'd like to get it installed in at least one or two popular addons prior to giving the talk. EPS was a great candidate because it is popular, already using ES classes, and the surface area of existing memory leaks seems limited. If ember-cli-memory-leak-detector gets traction from ember conf we can move the ember community toward eliminating memory leaks altogether!

## To-Do

That said, ember-cli-memory-leak-detector identified a few memory leaks that I haven't been able to fix on my own. I have narrowed down a few individual tests that exercise leaky code paths, however it's unclear whether it is the test itself that is leaky or the code that is being exercised that is leaking. 

One common theme I noticed is that tests that exercise touch events on options seemed to be causing leaks. I did find that in a few of these tests, just triggering the `touchstart` event on an option was enough to cause the PowerSelect class to be retained, but I couldn't identify any obvious leaks in the event handlers for touch events.

These tests can be isolated either via `ember test --filter="..."` or via `ember test --server` and manually filtering in the browser test page, and ember-cli-memory-leak-detector will either fail or pass individual tests based on the heap snapshot generated after running the individual test.
- [ ] [Integration | Component | Ember Power Select (Public actions) : If the options change and the new value is PromiseArrayProxy, the content of that proxy is set immediately while the promise resolves](https://github.com/cibernox/ember-power-select/blob/master/tests/integration/components/power-select/general-behaviour-test.js#L1185)
- [ ] [Integration | Component | Ember Power Select (Touch control) : Touch on option should select it](https://github.com/cibernox/ember-power-select/blob/master/tests/integration/components/power-select/touch-control-test.js#L24)
- [ ] [Integration | Component | Ember Power Select (Public actions) : The given `scrollTo` function is invoked when a single select wants to scroll to an element](https://github.com/cibernox/ember-power-select/blob/master/tests/integration/components/power-select/public-actions-test.js#L652)
- [ ] [Integration | Component | Ember Power Select (Public actions) : The given `scrollTo` function is invoked when a multiple select wants to scroll to an element](https://github.com/cibernox/ember-power-select/blob/master/tests/integration/components/power-select/public-actions-test.js#L671)


cibernox:
@steveszc sorry for delaying this so much, but finally I was able to find time to move CI to github actions and fix it.

Can you rebase, this time it should be green

cibernox:
@steveszc ping

steveszc:
@cibernox all rebased, however it looks like there are a couple memory leaks still. I've updated the PR description with the 3 tests I found that appear to be leaking the PowerSelect class. Haven't had time to dive into why they are leaking.

rwwagner90:
@steveszc revisiting this now. Would you possibly be able to rebase this PR?

steveszc:
@rwwagner90 Until this issue gets fixed it is unlikely that ember-cli-memory-leak-detector will run successfully with the newer versions of Qunit. I haven't had a chance to work on it, sadly.

PR #1443, opened 2021-03-25
add back customizing ember basic dropdown trigger and content components

betocantu93:
Sometimes reaching the core is needed, this was previously supported not sure when it got deprecated or regressed, i think when jumping from 3.0.6 to 4.x.x

`ember-paper` depends on it:

https://github.com/miguelcobain/ember-paper/blob/1c23f2ef65ffd3f496aab1ac1cc1ed88b3a5729f/addon/components/paper-select/template.hbs#L18

Previously:

https://github.com/cibernox/ember-power-select/blob/672ad84e7a96c4fdc8f1aff42b4c77be669e8156/addon/templates/components/power-select.hbs#L14

### DIFF
--- a/addon/components/power-select-multiple.hbs
+++ b/addon/components/power-select-multiple.hbs
@@ -54,7 +54,7 @@
   @verticalPosition={{@verticalPosition}}
   @tabindex={{this.computedTabIndex}}
   @ebdTriggerComponent={{@ebdTriggerComponent}}
-  @ebdTriggerComponent={{@ebdContentComponent}}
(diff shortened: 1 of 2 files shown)


cibernox:
I guess that sometimes it's inevitable for some abstractions to leak 😞 

Let's merge this once CI is green.

betocantu93:
Yea I know x.x , in my particular use case I actually need to completely replace the content, and so the whole power-select content renders as a Modal while basic-dropdown is open, for a mobile friendly select

cibernox (approved):

PR #1419, opened 2020-11-19
[Embroider] replace with safe component

akatz:
This replaces unsafe component usage with the embroider util that allows EBD to reach Optimized Embroider Safe support level as per https://github.com/embroider-build/embroider/blob/master/ADDON-AUTHOR-GUIDE.md

I also converted the documentation use because while the doc site will not need embroider with the static options enabled any time soon this does remove some deprecation warnings from the now unsafe string based api for passing components around.

PR #1463, opened 2021-08-18
Add full embroider compatibility

mydea:
This PR hopefully makes this addon fully embroider compatible.

Note that this relies on https://github.com/embroider-build/embroider/pull/922 (or something along these lines) to be merged - see https://github.com/embroider-build/embroider/issues/921 (in order for embroider compatibility to really work).

This is based on https://github.com/cibernox/ember-power-select/pull/1419 in many ways, but there have been a bunch of changes in the meanwhile, so it was somewhat easier to reproduce the changes than to deal with merge conflicts etc.

I also added some tests to ensure we keep supporting overwriting the default components (which would break with certain implementations, e.g. if we imported the default components in JS-land).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -90,7 +90,7 @@ jobs:
           - ember-beta
           # - ember-canary
           - node-tests
-          - embroider
+          - embroider-safe
           # - ember-classic
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -90,7 +90,7 @@ jobs:
           - ember-beta
           # - ember-canary
           - node-tests
-          # - embroider
+          - embroider
           # - ember-classic
     steps:
       - uses: actions/checkout@v2
--- a/addon/components/power-select-multiple.hbs
+++ b/addon/components/power-select-multiple.hbs
@@ -5,9 +5,9 @@
   @ariaInvalid={{@ariaInvalid}}
   @ariaLabel={{@ariaLabel}}
   @ariaLabelledBy={{@ariaLabelledBy}}
-  @afterOptionsComponent={{@afterOptionsComponent}}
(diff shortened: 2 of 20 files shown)


roomman:
Would love to see this progress. 👍🏻

I was using [this Webpack workaround](https://gist.github.com/bendemboski/907b8a53040213a911301ad48172b99b) but one of our projects uses `power-select-multiple` and that relies on dynamically setting `@triggerComponent`. 🤦🏻‍♂️

mydea:
I rebased this on current master and resolved any new conflicts, should be "good" again!

cibernox:
I managed to make CI a week ago, so if CI fails now I believe this time should be legitimate failures.

mydea:
OK, I think the failing tests should be fixed now :crossed_fingers:

cibernox:
@mydea I see, so in embroider-land one can't pass a component name to the select and have the select call `{{component @compName}}` on it. It has to be passed in as a component from the outside from the beginning.

I assume that this is not a breaking change as it's a requirement for embroider.

CI is green now! Merging it

cibernox:
I noticed that CI is not running embroider scenarios. I had to comment them because CI wouldn't pass. Can you add them back to `ember-try.js` and `.github/workflows/ci.yml`?

mydea:
> @mydea I see, so in embroider-land one can't pass a component name to the select and have the select call `{{component @compName}}` on it. It has to be passed in as a component from the outside from the beginning.
> 
> I assume that this is not a breaking change as it's a requirement for embroider.
> 
> CI is green now! Merging it

Yes, when passing something through like this you need to use `{{component (ensure-safe-component @arg)}}`, and it should be used by the user as `@arg={{component 'my-component'}}` instead of `@arg='my-component'`. The latter will still work in non-embroider land, so it should not be breaking (and I actually added a bunch of tests covering both scenarios). I'll need to see how I can make some of the tests optional in embroider, but that should be doable!

cibernox:
Great! Merged!

PR #1470, opened 2021-10-28
Multiselect: for disabled options, hide remove btn

toh995:
This is my first PR here, so please let me know if I'm missing anything!

For multiselects, if an option is disabled, AND that option is selected, then we shouldn't be able to "remove" that option from the selection. See videos below:

### Before
https://user-images.githubusercontent.com/52012721/139349835-7195a4f9-073f-455b-a4b6-068bb99ddb09.mov

### After
https://user-images.githubusercontent.com/52012721/139349765-f1a67204-e341-4b61-9058-96c783ec2569.mov


toh995:
Hello @cibernox are you the right person to tag to review this one?

cibernox:
I am. Checking it.

cibernox (approved):

cibernox:
It looks great! It was a clear oversight. Can you please add an entry on the Changelog describing this change? I'll merge it after that.

toh995:
Hi @cibernox thanks for the feedback!
I added a changelog entry - does it look correct?

cibernox:
Yes! Merging!

lougreenwood:
Just ran into this after upgrading to 5.1 - we now have failing tests and need to re-architect some UI. I wonder if this should be a breaking change?

For reference, our use case is a multi-select input where users list people to "invite", if a user adds an invalid email address we show an inline validation message and invite the user to delete the item (and disable the "submit" button until the invalid emails are removed).

With this change we need to not allow adding an invalid email to the multi select and come up with a new interaction which warns the user that the value they entered into the multi-select input is invalid and a new value should be entered...

toh995:
Yeah, seems fair to call this a breaking change 👍

cibernox:
I didn't anticipate this to be breaking, I'll undo it and consider options. 
As a base behavior, if an element is disabled it should not be possible to add or remove it. It's a reasonable default. But perhaps it should be configurable.

cibernox:
I reverted this in 5.0.2. I'll open a PR again with the revert of the rever and iterate there on a solution.

toh995:
Sounds good, thanks @cibernox!

lougreenwood:
Thanks @cibernox & @toh995.

Sorry for undercutting your first PR @toh995 - looking forward to seeing it land again and updating our code to work with it 😊.

toh995:
Hey @lougreenwood just seeing this now - no worries! Thanks for bringing this up

PR #1483, opened 2021-11-23
Use role=option for all options in list

calvin-fb:
I didn't include this in my previous PR (https://github.com/cibernox/ember-power-select/pull/1481) because I considered it a breaking change but since there is a plan to cut a new major release of EPS, I may as well try to sneak this in.

I haven't seen any documentation that using `role="alert"` is appropriate here (if such documentation exists, please send it to me!). The W3C reference implementation uses `role="option"` for all options (https://w3c.github.io/aria-practices/examples/combobox/combobox-select-only.html). In fact, our product had an accessibility audit completed by an external company and they said we had to switch it from `role="alert"`. As well, with the addition of `aria-activedescendant`, screen readers should be able to read out options without this.

We were able to patch this in our product by passing in a custom options component, but if I can sneak in this change, we can stop doing that.

### DIFF
--- a/addon/components/power-select/options.hbs
+++ b/addon/components/power-select/options.hbs
@@ -28,7 +28,7 @@
           aria-disabled={{if opt.disabled "true"}}
           aria-current="{{eq opt @select.highlighted}}"
           data-option-index="{{@groupIndex}}{{index}}"
-          role={{if (eq opt @select.highlighted) "alert" "option"}}>
+          role="option">
           {{yield opt @select}}
         </li>
       {{/if}}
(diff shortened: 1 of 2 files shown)


cibernox:
I missed this until now. Probably alert was only useful because with without activedescendant was the only way I had to have screen readers announced the selected option. If screenreaders properly speak the highlighted option when highlighting it then I'm good with removing this.

calvin-fb:
@cibernox Absolutely, that's my understanding as well. Give it a try with and without alert, I think it works better without it

cibernox:
Thanks! It works like a charm.

cibernox:
Released v5.0.0!

calvin-fb:
@cibernox Woohoo thanks!!!

PR #1484, opened 2021-11-23
aria-activedescendant property should not be present when dropdown is closed

calvin-fb:
The `aria-activedescendant` property should always refer to a valid element. When the dropdown is closed, none of the options are in the DOM, so `aria-activedescendant` should be blank.


cibernox (approved):

cibernox:
👍

PR #1495, opened 2022-01-05
Fix error when pressing backspace in empty multiple-select component

gvdp:
When pressing backspace in a multiple select component there was a console error logged because the selection is undefined: 

image

Fixed it by just adding a simple nullcheck in the trigger component.


rwwagner90:
@gvdp could you please rebase this PR?

gvdp:
@rwwagner90 Done. Thanks for having a look at it!

rwwagner90:
@gvdp looks like there are some linting issues. Mind running `lint:js --fix`?

gvdp:
@rwwagner90 My bad, didn't run the linting before committing. You can easily add a pre commit hook to prevent this from happening, I'll happily make a PR for that as well.

PR #1506, opened 2022-02-09
Cleans up scrollTo's element finding

matthew-robertson:
This commit simplifies the search for the option to scroll to, by querying directly for the element with the right index instead of fetching all elements and then selecting from the generated array.

This feels slightly cleaner to my eyes, and has the added benefit of ensuring it's not broken by something like [vertical-collection](https://github.com/html-next/vertical-collection), which only renders a sliding window of options for performance reasons. The combination was causing keyboard-navigation to fail unexpectedly, since `.item(19)` was no longer guaranteed to be the item with `data-option-index='19'`, nor was the array guaranteed to be long enough to have that element.

### DIFF
--- a/addon/components/power-select.ts
+++ b/addon/components/power-select.ts
@@ -430,7 +430,7 @@ export default class PowerSelect extends Component<PowerSelectArgs> {
     if (index === -1) {
       return;
     }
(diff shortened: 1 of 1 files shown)


calvin-fb (approved):

matthew-robertson:
Any thoughts on this @cibernox?

cibernox:
If all test pass, It's way cleaner. Thanks!

PR #1508, opened 2022-02-14
fix: staticComponents for consumer apps

betocantu93:
This PR allows consuming embroider addons/apps with `staticComponents` flag when using ember-power-select.

Notable changes
1. Update tests to stop using strings to reference components
2. Because of embroider's default ember-power-select [packageRules](https://github.com/embroider-build/embroider/blob/631eb5d1b273ac3dde54902fe9c074493f36bc1a/packages/compat/src/addon-dependency-rules/ember-power-select.ts#L21) we can't send dynamic components arguments to `<PowerSelect />` for any of those declared so, I moved the conditional to the JS side in order to just leave an static `{{component}}` helper . (see power-select-multiple.hbs @beforeOptionsComponent)
3. Unsure about this one but, everywhere I see in the ecosystem/guides the usage of (ensure-safe-component) is in conjunction with {{component}} helper and I also think the @embroider/compat `resolver-transform` expects this if its specified in packageRules

### DIFF
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -23,37 +23,5 @@ module.exports = function(defaults) {
     behave. You most likely want to be modifying `./index.js` or app's build file
   */
 
-  const safeDummyComponentsToIgnore = [
-    'custom-trigger-with-search',
-    'selected-country',
-    'custom-before-options',
-    'list-of-countries',
-    'custom-after-options',
-    'custom-search-message',
-    'custom-no-matches-message',
-    'custom-placeholder',
-    'custom-group-component',
-  ].reduce((acum, curr) => {
-    return {
-      ...acum,
-      [`{{component '${curr}'}}`]: {
-        safeToIgnore: true,
-      },
-    };
-  }, {});
-
-  const { maybeEmbroider } = require('@embroider/test-setup');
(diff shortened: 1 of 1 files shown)


addon/components/power-select-multiple.ts
```diff
@@ -25,7 +24,7 @@ export default class PowerSelectMultiple extends Component<PowerSelectMultipleAr
   }
 
   get triggerComponent() {
-    return this.args.triggerComponent || PowerSelectTriggerComponent;
+    return this.args.triggerComponent || 'power-select-multiple/trigger';
```
ef4, line 27: No, this is definitely a regression. Embroider prefers the class here.

betocantu93: Yes, I had it with the class but this [test](https://github.com/cibernox/ember-power-select/blob/c2f917e1b03aef0b85a0a09b12f9e05e9b04749d/tests/integration/components/power-select/customization-with-extending-components-test.js#L253) failed not sure how to handle this case

betocantu93:
Thanks for the review, yes I certainly got mislead by those rules... I guess the older versions means the ones without ensure-safe-component, gonna limit it with `semverRange`

PR #1512, opened 2022-03-02
Fix power-select with closeOnSelect=false nested within basic-dropdown

jamescdavis:
It turns out overriding `aria-owns` causes issues when a power-select is nested within a basic-dropdown and `@closeOnSelect={{false}}` is set on the power-select. Specifically, the parent dropdown is erroneously closed when an option is selected. This change was introduced in https://github.com/cibernox/ember-power-select/commit/0d9689088a238112b3749ebbb2b965ab15b9bac1 as part of #1481.

Repro: https://github.com/jamescdavis/power-select-nested-dropdown-repro/blob/main/app/templates/application.hbs

This PR removes that override and updates the test to look for the expected dropdown ID pattern (first commit).

While I was here, I updated all tests to use [hasAria](https://github.com/simplabs/qunit-dom/blob/master/API.md#hasaria) and [doesNotHaveAria](https://github.com/simplabs/qunit-dom/blob/master/API.md#doesnothavearia) (second commit).


jamescdavis:
@calvin-fb any ideas on a proper fix?

jamescdavis:
@matthew-robertson tagging you as well

calvin-fb:
@jamescdavis Interesting. I just played around with it a bit and I found that if I pass `@renderInPlace={{true}}` for the power-select, it also fixes the issue. If you are able to do that with your app, that would be a quick workaround until this gets fixed properly. I'm going to do some digging to find out what is actually going on and if there's a cleaner fix.

calvin-fb:
So the issue is with this block in ember-basic-dropdown:
https://github.com/cibernox/ember-basic-dropdown/blob/master/addon/components/basic-dropdown-content.ts#L390-L392
It expects `aria-controls` to be matched up with `ember-basic-dropdown-content`. With the accessibility fixes that I made, that's no longer the case. `aria-controls` is now connected to `ember-power-select-options` instead of `ember-basic-dropdown-content`.
Here's another relevant line:
https://github.com/cibernox/ember-basic-dropdown/blob/master/addon/components/basic-dropdown-content.ts#L352

In terms of fixing it... one hacky fix would be adding `ember-basic-dropdown-content` at the same level as `ember-power-select-options`. This is pretty gross though because the `ember-basic-dropdown-content` class would also be on the parent and if there any styling associated with it, it would look pretty gross.
The other option would be making a change in the ember-basic-dropdown repo to make it aware of the differences in ember-power-select (not ideal). The closestContent function would have to be changed to look for either `ember-basic-dropdown-content` or `ember-power-select-options`.

PR #1531, opened 2022-08-02
Update linting config, comply with prettier

rwwagner90:
This PR formats our code with prettier, in alignment with what the ember-cli blueprints do these days.


rwwagner90:
Going to merge this and follow up with additional embroider fixes

PR #1534, opened 2022-08-04
add back ember-try scenarios that were removed with ember-cli-update

mansona:
In https://github.com/cibernox/ember-power-select/pull/1533 the default blueprint update just removes the older ember versions from ember-try. If there is no reason to remove them then we can keep testing them and not drop support until we need to 👍 

Also 3.16 is a good base version to support because it's the first Octane LTS so there are no blockers to fully moving everything to Octane (if they haven't already been 😂 )

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,6 +61,8 @@ jobs:
       fail-fast: false
       matrix:
         try-scenario:
+          - ember-lts-3.16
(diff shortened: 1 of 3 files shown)


rwwagner90 (approved):

rwwagner90:
I'm fine adding these back for now, but if bugs arise in earlier Ember versions, I am not sure if we should fix them.

PR #1538, opened 2022-09-02
Add pre commit hook to run linting

gvdp:
Because I forgot to run the linting before committing in https://github.com/cibernox/ember-power-select/pull/1495#issuecomment-1235302705
This way it's not possible to commit something for which the linting fails.


tehhowch:
Why pre-commit, rather than pre-push?

rwwagner90:
Sorry, I do not agree with adding these hooks. Anyone should be able to commit, push, and open PRs with failing lint. They can then see lint is failing in CI.

PR #1540, opened 2022-09-07
be more permissive on ember-truth-helpers

miguelcobain:
Being more permissive on `ember-truth-helpers` helps people migrate their apps/addons.

For example, even ember-basic-dropdown that this addon depends on, does that: https://github.com/cibernox/ember-basic-dropdown/blob/master/package.json#L112

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,7 @@
     "ember-concurrency": ">=1.0.0 <3",
     "ember-concurrency-decorators": "^2.0.0",
     "ember-text-measurer": "^0.6.0",
-    "ember-truth-helpers": "^3.0.0"
+    "ember-truth-helpers": "^2.1.0 || ^3.0.0"
   },
   "devDependencies": {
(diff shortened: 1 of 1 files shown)


rwwagner90:
@miguelcobain could you please explain the need for this? The only breaking changes in ember-truth-helpers 3.x are dropping node and Ember versions, so I don't think we need to support older ember-truth-helpers here?

miguelcobain:
@rwwagner90 The thing is that, in some circumstances your project might have another dependency that uses ember-truth-helpers 2.x.
Because ember-power-select forces the use of 3.x, you end up with two versions on ember-truth-helpers installed, which makes https://github.com/salsify/ember-cli-dependency-lint not happy.

This change seems harmless, precisely because there are no breaking changes.

cibernox:
I think this is positive to avoid duplicated versions of ember-truth-helpers. It happened to me before to have duplicated dependencies just because an addon kept requiring the old version.

PR #1548, opened 2022-09-29
Add type declarations

charlesfries:
- Adds ambient type declarations for `selectChoose` and `selectSearch` test support functions.
- Exports the `Group` data structure interface as a convenience.


cibernox:
Thanks!

PR #1560, opened 2022-12-16
Fix import from ember-basic-dropdown

wagenet:
One should not import from the addon directory. In glint 1.0, this will cause problems if not addressed.


cibernox (approved):

cibernox:
thanks! And I didn't know of the existence of glint until 5 minutes ago!

wagenet:
@cibernox Hop on over to #topic-typescript on the Ember Discord if you want any pointers on first-class Glint support for your addons :) I'd be happy to help.

Issue #1552, opened 2022-11-11
Drop ember-concurrency-decorators


Yelinz:
ember-concurrency-decorators causes babel issues, decorators are now built-in in ember-concurrency

```
Build Error (broccoli-persistent-filter:Babel > [Babel: ember-concurrency-decorators]) in ember-concurrency-decorators/test-support/.eslintrc.js

[BABEL]: Cannot find module '@babel/helper-get-function-arity'
```

PR #1562, opened 2023-01-18
remove concurrency-decorators and update

BoussonKarel:
Drop ember-concurrency-decorators #1552


mkszepp:
@cibernox we can close this issue, because while updating all depenencies there was also removed `ember-concurrency-decorators`

PR #1570, opened 2023-03-06
Update to ember 4.8 & packages

mkszepp:
fix #1569 & #1563

PR #1572, opened 2023-03-06
Fix tests for embroider-optimized

mkszepp:
related to PR https://github.com/cibernox/ember-power-select/pull/1570


mkszepp:
@cibernox this shoud be the fix for the tests

cibernox:
Indeed it was! Thanks!

PR #1577, opened 2023-06-02
Update type imports -- /addon/ isn't a valid path

NullVoxPopuli:
for consumers that end up importing this file, the type declarations don't include `/addon`

`/addon` ends up working during development time, because the original source is in TS -- but it ends up not being the same as the emitted declarations.

### DIFF
--- a/addon/components/power-select.ts
+++ b/addon/components/power-select.ts
@@ -20,7 +20,7 @@ import {
 } from '../utils/group-utils';
 // @ts-ignore
 import { restartableTask, timeout } from 'ember-concurrency';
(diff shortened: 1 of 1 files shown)


wagenet (approved):
I've been using this change for a while in a fork of mine. I'd be happy to see it upstreamed.

cibernox:
This looks good. I believe you only need to rebase over master to get CI to pass

patricklx:
can this be fixed?

RobbieTheWagner:
Closed by #1589

PR #1578, opened 2023-06-04
restrict isSliceable to avoid slicing strings

mwhitworth:
I think the intention here is to `slice` arrays, rather than strings. Arrays can be sorted, but it's not an operation that makes sense for strings.

This caused a regression for a internal object that inherited from `String`. I would perhaps classify it as surprising behaviour.

### DIFF
--- a/addon/components/power-select.ts
+++ b/addon/components/power-select.ts
@@ -90,7 +90,7 @@ export interface PowerSelectArgs {
 }
 
 const isSliceable = <T>(coll: any): coll is Sliceable<T> => {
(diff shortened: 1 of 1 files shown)


cibernox:
Indeed, this was never meant for strings. If you could please rebase this branch I think that CI will be green

bertdeblock:
This change seems breaking? It's not compatible with `EmberArray`.
`EmberArray` does not have a `sort` method, it does have a `sortBy` method.

betocantu93:
it is breaking!

PR #1591, opened 2023-08-15
build: Fix embroider optimized CI builds

HeroicEric:
This fixes the `embroiderOptimized` ember-try scenario that runs as part of CI mostly by removing the dependency on other addons that cause issues:

- Removes `ember-cli-fastboot`
- Removes `ember-cli-deprecation-workflow`
- Removes `ember-fetch`

### Removing `ember-cli-fastboot`

This removes `ember-cli-fastboot` because it is not currently compatible with embroider optimized builds.

I believe `ember-cli-fastboot` was added as a dependency so `ember-power-select` could run some tests to ensure compatibility. This makes sense but AFAICT the tests to check fastboot compatibility have not actually been running since 2016. Since these fastboot tests have not actually been running for so long, IMO it makes sense to remove all of this for now and revisit fastboot tests separately.

Here is the commit where the command to run the fastboot tests was removed https://github.com/cibernox/ember-power-select/commit/1b9f90d123e95e4abfc72974946eb609a2149156#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L13

### Removing `ember-cli-deprecation-workflow`

There's [an issue with ember-cli-deprecation-workflow](https://github.com/mixonic/ember-cli-deprecation-workflow/issues/133) that causes an error during CI for the embroider optimized build.

I think we can remove `ember-cli-deprecation-workflow` from this addon because it does not seem to be doing anything useful but maybe I'm missing something?

This was originally added in https://github.com/cibernox/ember-power-select/commit/c10dd8eb5dc89a439541dac08fedd7878f9d544d as part of https://github.com/cibernox/ember-power-select/pull/1431 but there is not really much context to go off of as far as why it was added. It seems like maybe it was added while debugging GitHub Actions and accidentally left laying around?

- https://github.com/mixonic/ember-cli-deprecation-workflow/issues/133
- https://github.com/embroider-build/embroider/tree/main/packages/test-setup#embroideroptimizedembertryconfig

### Removing `ember-fetch`

This also removes the direct dependency on `ember-fetch` because it was only used in the docs/tests and causes some issues related to embroider.

The issue was reported here https://github.com/ember-cli/ember-fetch/issues/622 and the fix was basically to get around the issue by using native promises rather than RSVP. This root of the issue was not actually fixed yet and here is the open issue https://github.com/embroider-build/embroider/issues/1195


cibernox:
is this ready?

HeroicEric:
@cibernox should be ready now

cibernox:
thanks!

PR #1596, opened 2023-09-29
Add missing type for renderInPlace

cloke:
Added the missing type for renderInPlace which was causing gts templates to give an error. Added missing semi-colons to the interface since I was editing it anyway.


mkszepp:
@cloke in version v8 (atm beta) we have added all missing types, so we can close this PR

PR #1597, opened 2023-10-04
Avoid chained accessors in `.get()`

pkarman:
Ember 3.x deprecated the older style of `.get()` and some chained accessor paths have proven problematic for us.


mkszepp:
closing PR, because is select v2, we have already v7 released and in next time it comes v8 (code is already refactored in newer versions)

PR #1602, opened 2023-10-16
avoid passing arguments to the component `willDestroy`

### LINKED ISSUE #1601: TS2345 in a fresh ember app
Hey 👋 I've just noticed in a fresh ember typescript app with the latest(`7.1.2`) ember-power-select installed, I have the following TS error:

```sh
npm run lint:types 

> ln-ember-form@0.0.0 lint:types
> tsc --noEmit

../../node_modules/ember-power-select/addon/components/power-select.ts:148:35 - error TS2345: Argument of type 'IArguments' is not assignable to parameter of type '[]'.

148     super.willDestroy.apply(this, arguments);
                                      ~~~~~~~~~


Found 1 error in ../../node_modules/ember-power-select/addon/components/power-select.ts:148
```

This comes from the https://github.com/cibernox/ember-power-select/blob/ae707d0e4a0baf8477d2629a51a27cee00753648/addon/components/power-select.ts#L148, which has been introduced in https://github.com/cibernox/ember-power-select/pull/1392/commits/637354e054c44a06e08bdfc58d432d96bbc43937#diff-3a81f677e664ccdc7a0222ae6e84cdb47839fea70fcc92e6db2ad93a7d912badR141 and seems to be an accidental `arguments` usage here.

AFAIK, glimmer component `willDestroy` does not accept any arguments, so I believe we should be safe just to remove the invocation arguments. Thanks!

ro0gr:
fixes #1601

Unfortunately, I didn't manage to make the build fail even by upgrading TS to the latest. However, the change seems reasonable to me, since `willDestroy` isn't supposed to accept any arguments.

Знімок екрана 2023-10-16 о 15 10 14

### DIFF
--- a/addon/components/power-select.ts
+++ b/addon/components/power-select.ts
@@ -145,7 +145,8 @@ export default class PowerSelect extends Component<PowerSelectArgs> {
       } catch {}
       this._lastSelectedPromise = undefined;
     }
-    super.willDestroy.apply(this, arguments);
+
+    super.willDestroy();
   }
 
   // Getters


mkszepp:
@ro0gr i will close this issue, because it was fixed fix while updating the addon to an v2 addon (fix is released in `v8.0.0-beta.0`)

PR #1603, opened 2023-10-20
isSliceable is not working for ember data 3.x hasmany relationships

betocantu93:
This commit https://github.com/cibernox/ember-power-select/commit/83720b4d27abb4f04a4527769056e8e8fd23df79  introduced a bug for ED 3.27.X users (or earlier versions I guess, idk). HasMany relationships don't implement `.sort`, I think the idea is just to find out if the thing we are dealing with is an Arrayish, I would favor isArray util... which takes into account other ember array-like things


mkszepp:
Thank you for reporting! Yes i think this should fix the problem bug with ember-data 3.x. For ember-data 4.x & ember 5 there should no side effects.

I think we should add a test with ember-data for 3.28, while we support this old LTS

PR #1606, opened 2023-10-30
Change type from private to protected to enable inheritance

betocantu93:
When you want inheritance, accessing inputFont sometimes is necessary


mkszepp:
Hi @betocantu93,

i have removed this property completly in #1641, because there was a reported styling bug and i think this styling can also added directly in any css

Issue #1585, opened 2023-07-08
Types don't include Blocks or Element


ef4:
As of [RFC 748](https://github.com/emberjs/rfcs/blob/master/text/0748-glimmer-component-signature.md) components can declare the blocks they accept and the element they expose, if any.

People who are typechecking their templates will get type errors if they try to use power select and it doesn't declare these things, since power select does need to accept a block (and can accept html attributes).

A minimal version of this change would look something like:

```diff
-export default class PowerSelect extends Component<PowerSelectArgs> {
+export default class PowerSelect extends Component<{ Args: PowerSelectArgs, Blocks: { default: [any] }, Element: HTMLElement }> {
```

The `any` isn't ideal, but it's consistent with how the existing PowerSelectArgs are defined (which is that the user's model type is `any`). A larger change that would address that would make the whole component generic over the type of the user's model, so that `@options` is a `T[]`, `@selected` is a `T`, the default block yields a `T`, and so on.

mkszepp:
@ef4 this should be fixed with next version. While update the addon to v2 we have added missing properties and with latest 2 PR we have fixed the rest.

PR #1630, opened 2024-01-06
types: Separate Args and Signature

johanrd:
If`Element` and `Blocks` is included in the main component interface, `Args` needs its own property.

I had the following glint 2345 error when consuming the PowerSelect component, as it expected Element and Blocks to be passed in as required arguments:

```
NamedArgs<PowerSelectArgs> (…) is missing the following properties from type 'PowerSelectArgs': Element, destination, Blocks glint(2345)
```


johanrd:
fixes https://github.com/cibernox/ember-power-select/issues/1585

mkszepp:
Thank you!

PR #1649, opened 2024-01-17
Set `args.options` as optional

johanrd:
To me, all cases of undefined args.options seems to handled well, so undefined options should not give a glint type error.


mkszepp:
@johanrd yes its correct... i'm not sure, but i think we can also set `selected` nullable? (its the last property which is not nullable)

johanrd:
@mkszepp yes, good point. undefined args.selected seems to be handled well as well

PR #1650, opened 2024-01-17
Set `args.selected` as optional

johanrd:
See #1649. All cases of undefined seems to be handled well


mkszepp:
@johanrd can you resolve merge conflict? so i will merge it

PR #1690, opened 2024-02-12
Improve a11y & make easier to fullfill all requirements out of the box

mkszepp:
Some parts of a11y are not correct / not implemented since yet.

The goal is to make power-select out of the box ready for a11y by using [ARIA 1.2 specs](https://www.w3.org/TR/wai-aria-1.2/).

Changes:
- Add `@labelText`
  This adds `<label>` with all necessary attributes, that all a11y requirements are fulfilled
- Add `@labelClass` (allows to set styling classes on label tag)
- Add `@labelClickAction` (Default: focus)
  Click on label sets focus on power-select. This is the default behavior when using label-Tag with `<select>`. There is possible to pass `open` and the dropdown will be opened on label-tag click. This works only when using the new `labelText` paramenter
- Add `@labelComponent` to override standard label tag. Component get all necessary paramenter to make the same like we do internal in label component.
- Add `@resultCountMessage` (function). Allows to override default text which will only be used for readers. People which are using readers can here that the select has [xy] results. By searching this text will be updated.
  Default Text is [xy] results.

After changes we pass all Accessibility (tested with Lighthouse & some free tools):
grafik

Our docs page are passing now all requirements for a11y (there were necessary to modify some colors a little bit)

### DIFF
--- a/docs/app/components/snippets/action-handling-1.hbs
+++ b/docs/app/components/snippets/action-handling-1.hbs
@@ -1,3 +1,3 @@
-<PowerSelect @selected={{this.destination}} @options={{this.cities}} @onChange={{this.chooseDestination}} as |name|>
+<PowerSelect @selected={{this.destination}} @options={{this.cities}} @onChange={{this.chooseDestination}} @labelText="Country" as |name|>
   {{name}}
 </PowerSelect>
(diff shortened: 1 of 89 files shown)


mkszepp:
Discoverd that there are some braking part for power-select groups

PR #1714, opened 2024-02-27
fix: `Type error: (n.selected || []).slice is not a function` in multiple select

johanrd:
I had an issue popping up in sentry, `Type error: (n.selected || []).slice is not a function` here:  

```ts
defaultBuildSelection(option: any, select: Select) {
    const newSelection = (select.selected || []).slice(0);
````

The proposed fix checks for `Array.isArray(select.selected)` in runtime (which should not hurt), but we should probably also fix this by narrowing the type of Select.selected – possibly as a generic of what you pass in to options.


mkszepp:
Thank you!

PR #1715, opened 2024-02-29
Fix `NoMatchesMessage` component missing splattributes

alex-ju:
### Summary

In the `<PowerSelect>` component we call `<NoMatchesMessage>` [with attributes](https://github.com/cibernox/ember-power-select/blob/master/ember-power-select/src/components/power-select.hbs#L171-L173), but the `<NoMatchesMessage>` component is missing splattributes so they don't get rendered.

### Example

The example below uses the `helpers-testing-single-power-select` instance in the `test-app` with either a `@labelText="Label"` or an `@ariaLabel="Label"` provided to the `<PowerSelect>`.

#### Automated accessibility checks using [axe](https://github.com/dequelabs/axe-core)

<table>
<tr><td>

01-NoMatchesMessage-before

</td><td>

02-NoMatchesMessage-after

</td></tr>
</table>

#### Produced markup

Before

```html
<ul class="ember-power-select-options" role="listbox" id="ember-power-select-options-ember74">
    <li class="ember-power-select-option ember-power-select-option--no-matches-message" role="option">
      No results found
    </li>
</ul>
```

After

```html
<ul class="ember-power-select-options" role="listbox" id="ember-power-select-options-ember74" aria-label="Label">
    <li class="ember-power-select-option ember-power-select-option--no-matches-message" role="option">
      No results found
    </li>
</ul>
```

### Motivation

At @hashicorp we use `<PowerSelect>` extensively and we are grateful to see the many improvements introduced lately.

This change is primarily motivated by the desire to further improve the accessibility compliance of this component – in this instance, by always providing an accessible name to the `listbox` element (even when there are no results).

### DIFF
--- a/ember-power-select/src/components/power-select/no-matches-message.hbs
+++ b/ember-power-select/src/components/power-select/no-matches-message.hbs
@@ -1,5 +1,5 @@
 {{#if @noMatchesMessage}}
-  <ul class="ember-power-select-options" role="listbox">
+  <ul class="ember-power-select-options" role="listbox" ...attributes>
     <li class="ember-power-select-option ember-power-select-option--no-matches-message" role="option" aria-selected={{false}}>
       {{@noMatchesMessage}}
     </li>


mkszepp (approved):
Thans LGTM

PR #1716, opened 2024-02-29
Fix `BeforeOptions` search input missing `aria-expanded`

alex-ju:
### Summary

The search input in `BeforeOptions` has a `role="combobox"` which [requires an "aria-expanded" attribute](https://www.w3.org/TR/wai-aria-1.1/#combobox:~:text=Authors%20MUST%20set%20aria%2Dexpanded%20to%20true%20on%20an%20element%20with%20role%20combobox%20when%20it%20is%20expanded%20and%20false%20when%20it%20is%20collapsed.) (set to "true" or "false") depending on the state.

Generally, the `BeforeOptions` is only visible when the PowerSelect is expanded, but since we have access to its state we can use `@select.isOpen` to conditionally apply this value.

### Example

The example below uses the `helpers-testing-single-power-select` instance in the `test-app` with either a `@labelText="Label"` or an `@ariaLabel="Label"` provided to the `<PowerSelect>`.

#### Automated accessibility checks using [axe](https://github.com/dequelabs/axe-core)

<table>
<tr><td>

03-BeforeOptions-before

</td><td>

04-BeforeOptions-after

</td></tr>
</table>

### Motivation

At @hashicorp we use `<PowerSelect>` extensively and we are grateful to see the many improvements introduced lately.

This change is primarily motivated by the desire to further improve the accessibility compliance of this component – in this instance, by providing the required aria attribute for the combobox element.

### DIFF
--- a/ember-power-select/src/components/power-select/before-options.hbs
+++ b/ember-power-select/src/components/power-select/before-options.hbs
@@ -14,6 +14,7 @@
       aria-owns={{@listboxId}}
       aria-autocomplete="list"
       aria-haspopup="listbox"
+      aria-expanded={{if @select.isOpen "true" "false"}}
       placeholder={{@searchPlaceholder}}
       aria-label={{@ariaLabel}}
       aria-labelledby={{@ariaLabelledBy}}


mkszepp:
Thank you so much!

PR #1745, opened 2024-03-17
Make array arguments immutable

bendemboski:
Type the array component arguments as immutable. The component doesn't actually mutate them, so typing them as readonly allows caller to pass in either mutable or immutable arrays, rather than only mutable ones.


bendemboski:
Closing in favor of #1746

PR #1752, opened 2024-03-19
Consistent search experience

alex-ju:
### Summary

In this PR we propose a consistent search experience across the two main variations (single-selection and multi-selection). While I appreciate this would create a breaking change, and comes with tradeoffs, I believe it's worth exploring this not only for a unified search experience but also to reduce the code and maintainability required for two search experiences.

If maintainers are interested in considering this approach I can investigate the implications of such change more exhaustively.

#### Current experience with `@searchEnabled={{true}}`

<table>
<tr><th>Single-selection</th><th>Multi-selection</th></tr>
<tr><td>

01 power-select-search

</td><td>

02 power-select-multiple-before

</td></tr>
</table>

#### Proposed experience with `@searchEnabled={{true}}`

<table>
<tr><th>Single-selection</th><th>Multi-selection</th></tr>
<tr><td>

01 power-select-search

</td><td>

04 power-select-multiple-after

</td></tr>
</table>

### Examples

The example below uses the `helpers-testing-single-power-select` instance in the `test-app`.

<table>
<tr><td>

02 power-select-multiple-before

</td><td>

04 power-select-multiple-after

</td></tr>
<tr><td>

03 power-select-multiple-before

</td><td>

05 power-select-multiple-after

</td></tr>
</table>


mkszepp:
Hi @alex-ju thanks for this draft... i think it very good to align single & multiple. As you have described this is braking.

What do you thing about making an option like `@searchFieldPosition` (inline | content).
In v8 we can set for the single dropdown `@searchFieldPosition = 'content'` as default and for multiple `@searchFieldPosition = 'inline'`.

So we can bring this feature already in v8 and in next major update we do align this option.

alex-ju:
> Hi @alex-ju thanks for this draft... i think it very good to align single & multiple. As you have described this is braking.
> 
> What do you thing about making an option like `@searchFieldPosition` (default | inline). In v8 we can set for the single dropdown `@searchFieldPosition = 'default'` as default and for multiple `@searchFieldPosition = 'inline'`.
> 
> So we can bring this feature already in v8 and in next major update we do align this option. Of course we need also to extend the single select field with `inline` option like in https://react-select.com/home#welcome

That sounds like a good plan for a transition – and we can decide later if we want to keep both or not. Thanks for the feedback – I will likely close this one and open one that introduces `@searchFieldPosition`

mkszepp:
great thank! For position values maybe we can find something better... maybe `before-options` / `trigger` or if you have a better idea, feel free to define

PR #1784, opened 2024-04-16
Add `@searchFieldPosition` argument

alex-ju:
### Summary

Follow-up on #1752, gradually introducing `@searchFieldPosition` (`after-options` | `trigger`).

### Description

`<PowerSelect />` instances will have `@searchFieldPosition` set to `before-options` by default.
`<PowerSelectMultiple />` instances will have `@searchFieldPosition` set to `trigger` by default with the option to change it to `before-options`.

`<PowerSelectMultiple @searchEnabled={{true}} @searchFieldPosition="before-options" />` instances will have a search input placed in the dropdown, before the options, acting exactly as it currently does in `<PowerSelect @searchEnabled={{true}} />`.

### DIFF
--- a/ember-power-select/src/components/power-select-multiple.hbs
+++ b/ember-power-select/src/components/power-select-multiple.hbs
@@ -11,7 +11,7 @@
   @labelComponent={{ensure-safe-component @labelComponent}}
   @afterOptionsComponent={{ensure-safe-component @afterOptionsComponent}}
   @allowClear={{@allowClear}}
-  @beforeOptionsComponent={{if @beforeOptionsComponent (ensure-safe-component @beforeOptionsComponent) null}}
+  @beforeOptionsComponent={{if @beforeOptionsComponent (ensure-safe-component @beforeOptionsComponent)}}
   @buildSelection={{or @buildSelection this.defaultBuildSelection}}
   @calculatePosition={{@calculatePosition}}
   @closeOnSelect={{@closeOnSelect}}
@@ -49,6 +49,7 @@
   @search={{@search}}
   @searchEnabled={{@searchEnabled}}
   @searchField={{@searchField}}
+  @searchFieldPosition={{(or @searchFieldPosition 'trigger')}}
   @searchMessage={{@searchMessage}}
   @searchMessageComponent={{@searchMessageComponent}}
   @searchPlaceholder={{@searchPlaceholder}}
--- a/ember-power-select/src/components/power-select-multiple/trigger.hbs
+++ b/ember-power-select/src/components/power-select-multiple/trigger.hbs
@@ -52,7 +52,7 @@
       </li>
     {{/if}}
   {{/each}}
-  {{#if @searchEnabled}}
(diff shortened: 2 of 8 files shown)


mkszepp:
@alex-ju thank you... i will look to find time in next days to review it

mkszepp:
@alex-ju i have tested passing `@searchFieldPosition` for multiple and single select.
Multiple works great, but the `@searchFieldPosition="trigger"` on single doesn't work atm.

Do you have time to add also that option for single select?

alex-ju:
> @alex-ju i have tested passing `@searchFieldPosition` for multiple and single select. Multiple works great, but the `@searchFieldPosition="trigger"` on single doesn't work atm.
> 
> Do you have time to add also that option for single select?

I had a start on the single select as well, but it requires a bit more attention (esp on the a11y and focus management) so I was hoping to address it in a follow-up PR.

mkszepp:
> I had a start on the single select as well, but it requires a bit more attention (esp on the a11y and focus management) so I was hoping to address it in a follow-up PR.

ah, understood...
i would like to add in same PR, because we need to add the new paramenter in docs, so maybe other peoples are reporting us that the new paramenter is not working.

alex-ju:
> > I had a start on the single select as well, but it requires a bit more attention (esp on the a11y and focus management) so I was hoping to address it in a follow-up PR.
> 
> ah, understood... i would like to add in same PR, because we need to add the new paramenter in docs, so maybe other peoples are reporting us that the new paramenter is not working.

makes sense, will carry on with this PR then

alex-ju:
@mkszepp an accessible implementation of `@searchFieldPosition="trigger"` on single selection turns out to be rather complex. would you consider gradually introducing all combinations, by first getting `@searchFieldPosition="after-options" for multiple selection?

mkszepp:
@MelSumner atm i'm a little bit busy, but i would like to bring this feature to finish line. I think the open task is to bring the same logic like @alex-ju did already for multiple, to single select... (i think handling a11y support correctly, was a blocking point for him)

I would like to introduce this feature in current major version (v8) and before we make braking changes, because for sass we need to do a braking change (moving to modern api, replace `@import` with `@use` and drop support for non dart-sass). This is planned for begin of next year, otherwise we block peoples for using modern api (in my app, we were moved to vanilla js to fix this issue for now)

Would you like to complete this feature? or @alex-ju do you have any time to bring this to finish line?

Otherwise let me know, so i will look to schedule time for this feature.

mkszepp:
i have taken the changes from this PR and added the same for single select field in PR #1864
Tests should be pass (when npm registry is again fixed and up... npm is down :( )

i will close this PR as we have everything in PR #1864

@alex-ju thank you again for your input & help for adding this feature

mkszepp:
@alex-ju @MelSumner released in `v8.6.0`

alex-ju:
> @alex-ju @MelSumner released in `v8.6.0`

awesome, thanks for getting this over the line 🙌

PR #1841, opened 2024-06-21
Add `ember-inflector` in docs to fix docs page (cause ember-data bug)

mkszepp:
The current `ember-data` version brakes docs page. We don't use any `ember-inflector` but `ember-data` brakes production build when there is not installed

This PR fix for the moment our docs build without downgrading `ember-data` (see https://github.com/emberjs/data/issues/9501)

The error occurs only in production build, `ember test` or also with `ember s` there works fine

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -98,6 +98,7 @@
     "ember-concurrency": "^4.0.2",
     "ember-data": "~5.3.7",
     "ember-fetch": "^8.1.2",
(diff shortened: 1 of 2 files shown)


runspired:
for any future wanderers, the bug here is that ember-inflector is installed as a dependency by ember-cli-mirage, but not included in production builds if ember-cli-mirage is the only thing that brings it as a dependency, which then breaks @embroider/macros.

PR #1865, opened 2024-12-18
Fix multiple select error `TypeError Cannot read properties of null (reading 'length')`

### LINKED ISSUE #1854: TypeError Cannot read properties of null (reading 'length') on `this.searchText.length > 0`
Hi. In sentry I have some reports on `Cannot read properties of null (reading 'length')` on `this.searchText.length > 0` on [get results()](https://github.com/cibernox/ember-power-select/blob/913c85ec82d5c6aeb80a7a3b9d9c21ca9613e900/ember-power-select/src/components/power-select.ts#L293-L297)

Not sure if I am stretching the usage here, but anyways, I think ember-power-select should be resilient towards this.

Two fixes I can think of:

1) Optional chaining on all consumers of this.searchText.length (i.e. `this.searchText?.length`)

```diff
  get results(): any[] {
-    if (this.searchText.length > 0) {
+    if (this.searchText?.length > 0) {
      if (this.args.search) {
        return toPlainArray(this._searchResult || this.options);
```

2) Nullish coalescing on all assignments to this.searchText to ensure it is not null or undefined?

```diff
 @action
  _search(term: string): void {
    if (this.searchText === term) return;
-    this.searchText = term;
+    this.searchText = term ?? '';
    if (!this.args.search) {
      this.lastSearchedText = term;
      this._resetHighlighted();
    }
  }
  ```

johanrd:
…fixes cases where the result of `get(lastSelection, this.args.searchField)` is not guaranteed to be a string

see https://github.com/cibernox/ember-power-select/issues/1854#issuecomment-2551863511

close #1854

### DIFF
--- a/ember-power-select/src/components/power-select-multiple/input.ts
+++ b/ember-power-select/src/components/power-select-multiple/input.ts
@@ -77,7 +77,7 @@ export default class PowerSelectMultipleInputComponent extends Component<PowerSe
               this.args.searchField,
             );
             this.args.select.actions.search(
-              get(lastSelection, this.args.searchField),
+              get(lastSelection, this.args.searchField) || '',
             );
           }
           this.args.select.actions.open(e);


johanrd:
I guess this could also be solved also by typescript only (probably need to do a lot of conversion from `any` to generics), but having a failsafe in javascript is good anyways, I think.

mkszepp:
thank you for debugging and fix

mkszepp (approved):

mkszepp:
Released in `v8.6.1`

PR #1874, opened 2025-05-07
Allow ember-concurrency task `Promise<void>` in @search arg

johanrd:
Type support for case where ember-concurrency task.perform is set as the search function.

restartableTask<unknown, (term: string) => Promise<void>>(asyncArrowTaskFn: (term: string) => Promise<void>): TaskForAsyncTaskFunction<unknown, (term: string) => Promise<void>> (+4 overloads)

```
updateSearch = restartableTask(async (term: string) => {
    await timeout(250)
    this.router.transitionTo({ queryParams: { projectQuery: term } })
  })

<PowerSelect
@search={{this.updateSearch.perform}}
/>
```

### DIFF
--- a/ember-power-select/src/components/power-select.ts
+++ b/ember-power-select/src/components/power-select.ts
@@ -123,7 +123,7 @@ export interface PowerSelectArgs {
   search?: (
     term: string,
     select: Select,
-  ) => readonly any[] | Promise<readonly any[]>;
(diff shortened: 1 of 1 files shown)


johanrd:
okay, i see the integration tests fail. Not sure how to deal with that atm, as the search action may not return anything itself, if just writing reactive state to the url.

PR #1876, opened 2025-05-07
POC: Yield option as generic type

johanrd:
This could of course may need to be extended to the internal uses of options, results, and maybe Select, but adding this PR as a POC.

The goal is to have the yielded `{{option}}` correctly typed as passed through from what you pass in to @options

### DIFF
--- a/ember-power-select/src/components/power-select-multiple.ts
+++ b/ember-power-select/src/components/power-select-multiple.ts
@@ -4,11 +4,11 @@ import { isEqual } from '@ember/utils';
 import type { PowerSelectSignature, Select } from './power-select';
 
 // eslint-disable-next-line @typescript-eslint/no-empty-object-type
-interface PowerSelectMultipleSignature<T> extends PowerSelectSignature<T> {
+interface PowerSelectMultipleSignature<T = unknown> extends PowerSelectSignature<T> {
   // any extra property for multiple selects?
 }
 
-export default class PowerSelectMultipleComponent<T> extends Component<PowerSelectMultipleSignature<T>> {
+export default class PowerSelectMultipleComponent<T = unknown> extends Component<PowerSelectMultipleSignature<T>> {
   get computedTabIndex() {
     if (this.args.triggerComponent === undefined && this.args.searchEnabled) {
       return '-1';


mkszepp:
hmm.. i think this will be a big challange 🤔... i have tried this locally ~ 2 months ago the same, but after 1h and running from error to error i have stopped... moving `options` to generic type is tricky 😅..., but early or later we should do this, because the `any` is terrible

mkszepp:
I will close this PR, when my PR for resolving all `any` types is finished... The addon is working without `any` types, but there are some open tasks and also converting all tests to TS to ensure that everything works fine. You can follow this PR
https://github.com/cibernox/ember-power-select/pull/1906

This changes will be land in next major to avoid issues in stable v8 and that people must fix in app incorrect types.

johanrd:
@mkszepp wow, great, thanks! sounds very good :tada:

johanrd:
great! i see there is good progress on the typescript branch! really exciting!

For anyone needing yielded glint types in the mean time: I have had success adding these two files in `/types/ember-power-select/components/`:


```ts
/// power-select-multiple.d.ts
import Component from '@glimmer/component';
import type { PowerSelectArgs as OriginalPowerSelectArgs, Select } from 'ember-power-select/components/power-select';

interface PowerSelectMultipleSignature<T> {
  Element: HTMLElement;
  Args: Omit<OriginalPowerSelectArgs, 'options' | 'selected' | 'onChange'> & {
    options?: readonly T[] | Promise<readonly T[]>;
    selected?: readonly T[] | undefined;
    onChange: (selection: T[], select: Select, event?: Event) => void;
  };
  Blocks: {
    default: [option: T, select: Select];
  };
}

export default class PowerSelectMultiple<T = unknown> extends Component<PowerSelectMultipleSignature<T>> {}

```


```ts
/// power-select.d.ts
import Component from '@glimmer/component';
import type { PowerSelectArgs as OriginalPowerSelectArgs, Select } from 'ember-power-select/components/power-select';

type ArrayElement<T> = T extends readonly (infer U)[] ? U : T extends Promise<readonly (infer U)[]> ? U : never;

interface PowerSelectSignature<T> {
  Element: HTMLElement;
  Args: Omit<OriginalPowerSelectArgs, 'options' | 'selected' | 'onChange'> & {
    options?: readonly T[] | Promise<readonly T[]>;
    selected?: T | undefined;
    onChange: (selection: T | undefined, select: Select, event?: Event) => void;
  };
  Blocks: {
    default: [option: T, select: Select];
  };
}

export default class PowerSelect<T = unknown> extends Component<PowerSelectSignature<T>> {}

```

mkszepp:
yes... the TS branch is ready to go... it was a huge of work to make tests/docs... typed and shipping deprecation warnings in current 8.x version. The goal was to make the migration to 9.x easier as possible. People can resolve nearly everything in 8.x before moving to 9.x (resolving typing issues will be the biggest challanges for app's)

Right now, i'm waiting for ember 6.10 release (should be happen in ~6 weeks), to understand whats going breaking and needs to be fixed in our 8.x version / removed in 9.x...

PR #1877, opened 2025-05-10
fix: drop undesirable peers

runspired:
- @glimmer/tracking is provided by ember-source
- ember-source is assumed by both embroider and ember-cli in ways that it doesn't actually need to be a peer, and having it as one creates potential for duplicate copies that create errors in apps


mkszepp:
Here more details

https://github.com/ember-cli/ember-addon-blueprint/pull/35

Issue #1936, opened 2025-08-21
Plans for v9 release


mkszepp:
- [x] Improve typescript story (remove `any` types) and make necessary changes #1906 
- [x] Drop support for ember < 4.12 (#1961)
- [x] Convert helpers to plain functions and drop `ember/component/helper` usages (works with ember 4.5+) (#1906)
- [x] Remove blueprints (#1962)
- [x] Drop support for old dependencies/peerDependencies (#1963)
- [x] Drop support for passing components as string (#1906)
- [x] Remove `@embroider/util` (deprecated and not anymore needed after dropping option to pass components as string & gts convert, since ember 3.25 its recommended to pass component) (#1960)
- [x] Convert addon to gts (#1960)
- [x] Add css variables support (#1965)
  - [x] Change behavior for `$ember-power-select-line-height` => pass always with unit
- [x] Remove automatically added ember-basic-dropdown (#1964)
- [x] Drop support for `node-sass` (#1964)
- [x] Fix scss deprecations (#1964)
- [x] Remove built-in CSS/SCSS themes for Bootstrap and Material (#1964)
- [x] Fix Accessibility issue in label component (#1966)
- [x] Move to glint v2 (#1968)
  - [x] Convert test-app to gts (#1967)
- [x] Remove `ember-assign-helper` dependency (#1969)
- [x] Migrate to new addon v2 blueprint https://github.com/ember-cli/ember-addon-blueprint (#1971)
- [x] Drop support for `ArrayProxy` (not recommended since 3.x, ember data has deprecated in 4.x cycle, makes possible to remove dependency of `addObserver`/`removeObserver`) (#1974)
- [x] Check, if there is possibile to drop dependency of  `@ember/utils` after `ArrayProxy` drop (use native ways, works better with TypeScript, it will be deprecated in Polaris) (#1974)
- [x] Remove added deprecations in v8.x (#1976)

Target date: tba

mkszepp:
v9.0.0 is out

PR #1951, opened 2025-11-11
Convert helpers to plain functions and drop ember/component/helper usages

kexposito:
Remove the `ember/component/helper` and use plain functions instead as is described [here](https://github.com/cibernox/ember-power-select/issues/1936) for v9

### DIFF
--- a/test-app/tests/unit/helpers/ember-power-select-is-equal-test.ts
+++ b/test-app/tests/unit/helpers/ember-power-select-is-equal-test.ts
@@ -1,0 +1,32 @@
+import { module, test } from 'qunit';
+import {
+  emberPowerSelectIsEqual,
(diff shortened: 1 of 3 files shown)


ember-power-select/src/helpers/ember-power-select-is-group.ts
```diff
@@ -1,8 +1,7 @@
-import { helper } from '@ember/component/helper';
 import { isGroup } from '../utils/group-utils.ts';
 
-export function emberPowerSelectIsGroup([maybeGroup]: [any]): boolean {
+export function emberPowerSelectIsGroup(maybeGroup: any): boolean {
```
kexposito: here we could just now call `isGroup` directly when we use it (just 1 place) and remove this helper. same with the `isNone`, let me know if we want to do that change.

PR #1971, opened 2025-12-31
Convert to new addon v2 blueprint

mkszepp:


bwbuchanan:
@mkszepp This missed the addition of the `ember-concurrency/async-arrow-task-transform` plugin to `babel.publish.config.cjs`, resulting in a crash when attempting to use the component.

PR #2047, opened 2026-05-11
A11y: fix incorrect handling on clear & selected item buttons

malatin3:
Hello!
We are having an accessibility issue where in cases where `@multiple=true` the user cannot drag the pointer off the selected item after clicking in order to cancel the removal of said item. Same issue exists for the clear button.

Believe this is an issue with the `on "mousedown"` usage. I fixed it by changing it to work off of pointerup and not allowing the "down" events to propagate when the user is clicking on a button itself.

Not sure if it's the most graceful solution. Would be happy to update PR if there is a better way and this is wanted.

Thank you

### DIFF
--- a/src/components/power-select/trigger.gts
+++ b/src/components/power-select/trigger.gts
@@ -157,7 +157,7 @@ export default class PowerSelectTrigger<
         }}
         class="ember-power-select-multiple-options"
         {{this.openChange @select.isOpen}}
-        {{on "pointerup" this.chooseOption}}
+        {{on "click" this.chooseOption}}
         ...attributes
       >
         {{#if (selectIsSelectedPresent this.selectedMultiple)}}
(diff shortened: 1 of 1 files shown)


mkszepp:
@malatin3 thanks for reporting this issue! Yea, `mousedown` events are bad for accessibility.
I have seen that tests were failed... this happens because in `qunit` the `pointerup` event isn't working. Maybe we can change to `click` event. This should fix the tests and also your issue.

malatin3:
@mkszepp I think the test issues should be resolved. 

I swapped it to click like you suggested. When I tested it on my phone it was not 100% behaving correctly, but adding an additional prevent default to the `touchend` seemed to make it happy 😅

mkszepp (approved):
Thank you!

PR #2084, opened 2026-07-28
Treat empty strings as an `undefined` selection

### LINKED ISSUE #2082: Falsy `@selected` values act differenly in v8+
I'm updating one of our apps from power-select v7 to v8. After the update some of our components started acting strangely. We have some search filters that are wired up to query params. They default to an empty string and we pass these to our PowerSelect component through the `@selected` value.

In v7 this worked fine. The select was in an empty state.

After the v8 update, however, PowerSelect thinks it's in a selected state and displays an "empty" option with the clear button next to it.

Image

I did some spelunking and I think this PR is the cause: https://github.com/ember-power-addons/ember-power-select/pull/1674 After that change power-select explicitly check for null and undefined instead of falsy, so empty strings behave differently as well.

Are empty strings supposed to work and is this a bug, or was it never intended to work the way we used it? If it's intentional an extra note in the upgrade guide might be nice.

Windvis:
In ember-power-select v7, empty strings were considered an "empty" value and the select would be in an unselected state.

In v8, an issue was fixed where the number `0` wasn't considered a valid select option, but the way it was implemented also unintentionally broke the empty string behavior.

This change restores the old empty string behavior while still supporting the number 0 as well.

Closes #2082

### DIFF
--- a/src/helpers/ember-power-select-is-selected-present.ts
+++ b/src/helpers/ember-power-select-is-selected-present.ts
@@ -1,5 +1,5 @@
 export default function emberPowerSelectIsSelectedPresent<T>(
   value: T | undefined,
 ): value is T {
-  return value !== null && value !== undefined;
+  return value !== null && value !== undefined && value !== '';
 }
--- a/tests/integration/components/power-select/general-behaviour-test.gts
+++ b/tests/integration/components/power-select/general-behaviour-test.gts
@@ -2222,6 +2222,39 @@ module(
         .hasText('fourteen');
     });
 
+    test<NumbersContext>('An empty-string selection is treated as an empty select', async function (assert) {
+      const self = this;
+
+      assert.expect(2);
+
+      this.numbers = numbers;
+      this.selected = '';
+      this.foo = () => {};
+
+      await render<NumbersContext>(
+        <template>
+          <HostWrapper>
+            <PowerSelect
+              @options={{self.numbers}}
+              @selected={{self.selected}}
+              @allowClear={{true}}
+              @onChange={{self.foo}}
+              as |option|
+            >
+              {{option}}
+            </PowerSelect>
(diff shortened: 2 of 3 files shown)


src/helpers/ember-power-select-is-selected-present.ts
Windvis, line 1: I reused and extended this helper, but not sure if a new util would be better. Are these helpers public API?

tests/integration/components/power-select/general-behaviour-test.gts
```diff
@@ -2222,6 +2222,39 @@ module(
...
+            >
+              {{option}}
+            </PowerSelect>
+          </HostWrapper>
+        </template>,
+      );
+
+      assert
+        .dom('.ember-power-select-trigger', getRootNode(this.element))
+        .hasText('', 'The trigger stays empty');
```
Windvis, line 2252: Bit of a silly test I guess. Not sure what the best way is to verify that the select is in an unselected state.

src/helpers/ember-power-select-is-selected-present.ts
mkszepp: i think is fine to take this helper, so the Template and JS part have the same behavior.
The helper is public (i think we have marked this public in v8.x), because some apps have copied always the component templates for making custom into there app.

mkszepp (approved):
Thank you!

mkszepp:
@Windvis Is the fix in v9 enough for you, or do you need it in v8 as well?

Windvis:
> @Windvis Is the fix in v9 enough for you, or do you need it in v8 as well?

@mkszepp v9 is fine. There is an easy workaround available :+1: Thanks for the quick review!

mkszepp:
@Windvis released in v9.0.2

